PR #2, opened 2021-05-21
meeting creation form fully functional

JakubC-projects:


client/src/util/churches.js
```diff
@@ -0,0 +1,226 @@
+export default [
```
philipdalen, line 1: I would say this is fine for now, but i suggest that in the future this gets refactored to get the list of churches directly from members.bcc.no ideally through a web components which could live in members. That way we could start a service oriented architecture, or what do you think @StevenMalaihollo

client/src/views/Create.vue
```diff
@@ -30,39 +30,55 @@
...
       </date-picker>
     </div>
-    <div class="horizontal padding-vertical-20">
-      <input type="checkbox" class="margin-horizontal-10" v-model="publicVoting"/>
-      <label class="medium-text">Public Results</label>
-    </div>
     <div class="horizontal width-100 space-out">
     <h2>Invited Voters</h2>
-    <h2 v-if="numOfVoters">Found {{numOfVoters}} Voters</h2>
+    <h2>Found {{numOfVoters}} People</h2>
```
philipdalen, line 35: Perhaps "Members" instead of "People" ?

client/src/util/churches.js
StevenMalaihollo: I will have to look into serving components from the server and if it is valuable. Styling and layout is often page specific.

server/src/services/members/members.class.ts
```diff
@@ -38,58 +38,39 @@ export class Members implements ServiceMethods<Data> {
...
+
+    url += "&churchID="+churchID;
+
+    // if(minAge)
+    //   url += "&age[$gt]="+minAge;
+
+    // if(maxAge)
+    //   url += "&age[$lt]="+maxAge;
+
+    console.log(url);
```
StevenMalaihollo, line 67: Remove all console logs and unneeded variabeles that we won't use in the future.

StevenMalaihollo (approved):

PR #38, opened 2021-05-26
Setup for vuex and typescript and linting

JakubC-projects:


client/src/router/routes.js
```diff
@@ -8,6 +9,11 @@ export const routes = [{
     name: 'Home',
     component: Home
   },
+  {
+    path: '/test',
+    name: 'Test',
```
StevenMalaihollo, line 14: Hmm, now we have a JS and a TS file for the same thing?

I think we should remove the JS files

JakubC-projects: Ok, I can remove them

StevenMalaihollo (approved):
This is good for now, we need to get this merged so we don't step on eachothers feet while developing

JakubC-projects (commented):
Good job

PR #52, opened 2021-05-28
CI/CD fully setup

philipdalen:
- Updated environment variable names by removing BCC_ (Please ask Philip for a his set)
- Fixed 401 issues by getting package from github, this works but you will have to get a token to you're .npmrc user file on your local machine
- Added support for ci/cd
- Made frontend more knowledgable about the domain it runs in
- Created a new client in Auth0 for the voting app


JakubC-projects (commented):
I don't know much about CI/CD, but I've been testing the website a bit (dev.vote.bcc.no) and it seems to be working fine. Though half an hour ago I would often get a bug when trying to log in: error=Grant: missing session or misconfigured provider. Then it somehow stopped showing and I could make 20 successful logins in a row.

JakubC-projects (approved):

PR #65, opened 2021-05-31
Poll adding component

JakubC-projects:


client/src/components/PollForm.vue
```diff
@@ -0,0 +1,97 @@
+<template>
```
StevenMalaihollo, line 1: Don't forget the component casing:

poll-form.vue

client/src/components/PollForm.vue
```diff
@@ -0,0 +1,97 @@
...
+    <div class="flex justify-center pt-8 pb-4">
+      <div class="gradient-button text-lg flex items-center cursor-pointer"  @click="createPoll">
+          <PlusIcon class="w-6 h-6 p-0.5 mr-2"/>
+          {{$t('actions.add-question')}}
+      </div >
+    </div>
+  </div>
+</template>
+
+<script>
```
StevenMalaihollo, line 36: I think we should build all components with the typescript engine.

Using defineComponent and adding types in our /domain folder

client/src/components/PollForm.vue
```diff
@@ -0,0 +1,97 @@
...
+
+export default {
+  components: {
+    InfoBox,
+    FormField,
+    PlusIcon,
+  },
+  props: {
+    pollingEventId: String,
+    index: Number,
```
StevenMalaihollo, line 51: index is maybe a bit too generic name? index of what? Is it the sequence on all polls?

StevenMalaihollo:
I think you have to merge Master into the branch

client/src/components/poll-form.vue
```diff
@@ -0,0 +1,98 @@
...
+        label: "",
+        explanation: "",
+        answerId: 2,
+      })
+    },
+    removeOption(ind: number){
+      this.pollData.answers.splice(ind, 1);
+    },
+    createPoll(){
+      this.$client.service('poll').create(this.pollData)
```
StevenMalaihollo, line 90: not needed?

client/src/components/poll-form.vue
```diff
@@ -0,0 +1,98 @@
+<template>
+  <div>
+    <InfoBox class="mb-8">
+      {{$t('info.question-visibility')}}
+    </InfoBox>
```
StevenMalaihollo, line 5: I think this info box should be in the parent component. This info box is specific to the Poll creation page so it doesn't fit in this component that we will reuse often

client/src/views/PollingEventCreate.vue
```diff
@@ -42,18 +42,28 @@
...
   </div>
 </template>
 
-<script>
+<script lang="ts">
 
-import InfoBox from '../components/info-box'
-import FormField from '../components/form-field'
-export default {
+interface church {
```
StevenMalaihollo, line 52: We store the Types in the domain folder and you can already use the one that is there :)

client/src/views/PollingEventCreate.vue
```diff
@@ -109,31 +119,37 @@ export default {
...
             return {
               name: c.name,
               val: c._key,
             }
           });
-          this.allRoles.unshift({name: 'All roles', val: 0})
       },
       createPollingEvent(){
-        const data = this.eventData;
+        const data:any = this.eventData;
```
StevenMalaihollo, line 137: Use the PollingEvent type here

JakubC-projects: How do I do that if in the data the id isn't yet defined

client/src/App.vue
```diff
@@ -4,9 +4,7 @@
     <template v-else>
       <div :class="['app-container',backgroundClass]">
         <Header />
-        <transition name="fade" mode="out-in">
-            <router-view :key="$route.fullPath" :style="`min-height: calc(100vh - ${layoutHeight}px);`"/>
-        </transition>
```
JakubC-projects, line 9: Those transitions don't work and make it so that you cannot refresh your page

PR #72, opened 2021-06-01
Implimented tests for polling-event

philipdalen:


client/src/domain/PollingEvent.ts
```diff
@@ -3,14 +3,28 @@ export interface PollingEvent {
     _key: string;
     title: string;
     description: string;
-    type: string;
+    type: PollingEventType;
```
StevenMalaihollo, line 6: These polling types break the polling-event-card component.

Look at how the types are referenced. They have translations attached which you can replace or refactor.

client/src/domain/PollingEvent.ts
```diff
@@ -3,14 +3,28 @@ export interface PollingEvent {
...
+    startDateTime: Date;
     creatorId: string;
     participantFilter: ParticpantFilters;
 }
 
 export interface ParticpantFilters {
-    orgs: Array<string>;
-    roles: Array<string>;
+    orgs: string
+    roles: string;
```
StevenMalaihollo, line 15: These types probably don't match the ones in the PollingEventCreate view. Could you check it out?

StevenMalaihollo (approved):

PR #76, opened 2021-06-01
Editing is possible

JakubC-projects:


client/src/assets/style.css
```diff
@@ -92,13 +92,15 @@
   .lg-button {
     @apply py-4;
     @apply px-6;
+    @apply cursor-pointer;
```
StevenMalaihollo, line 95: This styling doesn't fit here :/ 

lg / bg classes are for sizes and cursor is for hover state right. Good css is mostly about creating reusable blocks. So i think what you mean to do is to add the cursor-pointer class to all buttons elements or something like that.

client/src/components/poll-form.vue
```diff
@@ -1,31 +1,37 @@
 <template>
-  <div>
-    <h3 class="font-bold mb-6">{{pollIndex}}. {{$t('labels.poll')}}</h3>
+  <div class="border-with-bar px-5 py-8">
+    <div class="flex justify-between mb-6">
+    <h3 class="font-bold">
+      {{this.poll ? this.pollIndex+'. '+this.poll.title:$t('labels.new-poll')}}
```
StevenMalaihollo, line 5: this shouldnt be needed inside the html. Is this working?

client/src/components/poll-form.vue
```diff
@@ -105,14 +97,24 @@ export default defineComponent({
...
+        })
+      }
+      else{
+        this.$client.service('poll').create(this.pollData)
+        .then(() => {
+          this.$emit('close');
+        })
+      } 
+    },
+    changePoll(){
```
StevenMalaihollo, line 114: unneeded?

client/src/components/saved-poll.vue
```diff
@@ -1,34 +1,57 @@
...
+        this.isEdit = true;
+      }
+    },
+    stopEditing(){
+      this.$emit('stopEdit');
+      this.isEdit = false;
+    },
+    deletePoll(){
+      if(this.active)
+        this.$client.service('poll').remove(this.poll?._key)
```
StevenMalaihollo, line 52: This is not awaited ;)

StevenMalaihollo (approved):

server/src/channels.ts
```diff
@@ -25,4 +25,13 @@ export default function(app: Application): void {
...
   });
+
+
+  app.services.poll.publish('updated', (data:any) => {
+    return app.channel(data.pollingEventId);
+  });
+
+  app.services.poll.publish('removed', (data:any) => {
+    return app.channel(data.pollingEventId);
+  });
```
StevenMalaihollo, line 36: These are not used anymore right?

StevenMalaihollo (approved):

PR #81, opened 2021-06-02
Basic live connection

JakubC-projects:


client/src/views/AdminEventLive.vue
```diff
@@ -1,61 +1,111 @@
...
         InfoBox,
-        PencilIcon
+        PencilIcon,
+        PollForm,
+        SavedPoll,
     },
     data() {
       return {
         currentTab: 'polls' as string,
         pollingEvent: {
```
StevenMalaihollo, line 53: I think that this event item should be null until we get it from the server. not with all the props defined

client/src/views/AdminEventLive.vue
```diff
@@ -1,61 +1,111 @@
...
+      },
+      reloadPolls(){
+        this.currentlyEdited = 0;
+        this.loadSavedPolls();
+      },
+      startEditing(ind: number){
+        this.currentlyEdited = ind + 1; 
+      },
+      publishPoll(ind: number){
+        console.log(ind);
```
StevenMalaihollo, line 103: console logs

server/src/permissions/appAbility.ts
```diff
@@ -18,6 +18,8 @@ const globalPermissions = (user: UserDetails, { can, cannot }: AbilityBuilder<Ap
     can('update', 'poll');
     can('remove', 'poll');
     can('patch', 'polling-event');
+    can('patch', 'poll');
+    can('get', 'poll');
```
StevenMalaihollo, line 22: I think that getting a poll should be available for Non developers too

client/src/views/AdminEventLive.vue
```diff
@@ -1,61 +1,111 @@
...
+        })
+      },
+      reloadPolls(){
+        this.currentlyEdited = 0;
+        this.loadSavedPolls();
+      },
+      startEditing(ind: number){
+        this.currentlyEdited = ind + 1; 
+      },
+      publishPoll(ind: number){
```
StevenMalaihollo, line 102: I am not sure that Index is a safe way to filter out the active poll. Maybe use the _key

StevenMalaihollo (approved):

PR #89, opened 2021-06-02
Added form validation and changed to socket connection to members

philipdalen:


StevenMalaihollo (approved):

StevenMalaihollo (approved):

server/src/services/poll/poll.hooks.ts
```diff
@@ -1,16 +1,20 @@
...
   before: {
     all: [],
     find: [],
     get: [],
-    create: [],
-    update: [],
-    patch: [],
+    create: [validateAndFormat],
+    update: [validateAndFormat],
+    patch: [validateAndFormat],
```
JakubC-projects, line 17: I think we shouldn't validate on patch (or have some special validation), because data there doesn't contain all the fields - just the ones that are changed.

server/src/services/polling-event/polling-event.hooks.ts
```diff
@@ -1,8 +1,21 @@
...
+const { authenticate } = authentication.hooks;
 // Don't remove this comment. It's needed to format import lines nicely.
 
-const { authenticate } = authentication.hooks;
+
+
+const validateAndFormat = (context: HookContext) => {
+  const { data } = context;
+
+  if (data.participantFilter.maxAge == undefined ) data.participantFilter.maxAge = 150
```
StevenMalaihollo, line 11: This is causing a problem when the participantFilter is not patched

PR #90, opened 2021-06-03
Vote redirect to results after vote

StevenMalaihollo:


client/src/components/poll-popover-results.vue
```diff
@@ -0,0 +1,36 @@
+<template>
+    <div class="h-full w-full px-4 md:px-6 py-8 bg-white rounded-t-lg">
+        <div v-for="(option,index) in options" :key="option.answerId" class="mb-6">
+            <div class="w-full result-bar flex" :style="`background: linear-gradient(to right, ${answerColors[index]} ${answerPercent[index]}%, #FFF ${100 - answerPercent[index]}%) ;`">
```
JakubC-projects, line 4: Shouldn't the gradient be from answerPercent[index] to answerPercent[index], so that there is no gradient, because I don't think there was supposed to be some gradient

client/src/components/poll-popover.vue
```diff
@@ -1,45 +1,32 @@
...
         poll: Object as PropType<Poll>,
     },
     data() {
         return {
-            selectedOption: null
+            chosenAnswer: {} as Answer,
+            hasSavedAnswer: false as boolean
         }
     },
     methods: {
```
JakubC-projects, line 32: at created we should check if the user has already answered the poll

client/src/components/poll-popover-results.vue
StevenMalaihollo: True, will fix in next PR, this is just for showing proof of concept to management

PR #91, opened 2021-06-03
besic result live update

JakubC-projects:


client/src/components/poll-popover.vue
```diff
@@ -48,7 +48,8 @@ export default defineComponent({
                 const participantAnswer = {
                     _from: this.poll._id,
                     _to: this.$user._id,
-                    answerId: option.answerId
+                    answerId: option.answerId,
+                    pollingEventId: this.$route.params.id
```
StevenMalaihollo, line 52: The pollingEventId is an _id right? the _key is like '123213123'

client/src/views/AdminPollResult.vue
```diff
@@ -0,0 +1,46 @@
...
+      }
+    })
+    res.forEach((ans: Answer) => {
+      this.answers[this.answerMap[ans.answerId]] ++;
+    });
+    this.$client.service('answer').on('created', this.updateAnswers);
+  },
+  methods: {
+    updateAnswers(ans: Answer){
+      console.log('new ans');
```
StevenMalaihollo, line 41: console logs

client/src/views/AdminPollResult.vue
```diff
@@ -0,0 +1,46 @@
...
+  },
+  async created(){
+    this.poll = await this.$client.service('poll').get(this.$route.params.id)
+    let ind = 0;
+    this.poll.answers.forEach((ans: Answer) => {
+      this.answerMap[ans.answerId] = ind;
+      ind++;
+    });
+    this.answers = new Array<number>(ind).fill(0);
+    const res = await this.$client.service('answer').find({
```
StevenMalaihollo, line 28: It is a bit unclear what you are trying to do here. This can have to do that you are using dev language like ind and res as variabeles.

Could you maybe create some methods and rename some stuff?

server/src/channels.ts
```diff
@@ -11,6 +11,7 @@ export default function(app: Application): void {
     return app.channel(data._key);
   });
   app.services.answer.publish('created', (data:any) => {
+    console.log(data.pollingEventId);
```
StevenMalaihollo, line 14: console log

client/src/components/poll-popover.vue
JakubC-projects: the pollingEventId is actually the _key

server/src/services/poll/poll.hooks.ts
```diff
@@ -21,7 +27,7 @@ export default {
   after: {
     all: [],
     find: [],
-    get: [],
+    get: [ addChannel ],
```
StevenMalaihollo, line 30: Do you have a test for this functionality? Philip mentioned that all logic in the server needs to be confirmed by a test.

client/src/views/AdminPollResult.vue
```diff
@@ -0,0 +1,64 @@
...
+    async created(){
+        await this.loadPoll()
+        this.createAnswerMap()
+        await this.loadAnswers()
+
+        this.$client.service('answer').on('created', this.updateAnswers)
+    },
+    methods: {
+        async loadPoll(){
+            this.$client.service('poll').get(this.$route.params.id)
```
StevenMalaihollo, line 31: There is an error here:

Cannot read property 'forEach' of undefined in createAnswerMap

This is because the map  is created before the data is awaited. That is why we use the async - await combo instead on the .then() function.

StevenMalaihollo (approved):

PR #112, opened 2021-06-05
automatic reconnection + connection log in ui

JakubC-projects:


server/src/channels.ts
```diff
@@ -2,36 +2,44 @@ import '@feathersjs/transport-commons';
...
-
-  app.services.poll.publish('removed', (data:any) => {
-    return app.channel(data.pollingEventId);
-  });
+    if(typeof app.channel !== 'function') {
+        return;
+    }
+
+    app.on('login', (payload:any, { connection }:any) => {
+        const lastChannel = connection.user.lastChannel;
```
StevenMalaihollo, line 10: This is throwing an error upon login for me:

TypeError from app.hooks.ts: Cannot read property 'user' of undefined during create authentication

JakubC-projects: whoops I forgot to check if there is a connection present

client/src/components/log-information.vue
```diff
@@ -0,0 +1,74 @@
+<template>
+    <div class="absolute log-container p-2 bg-black opacity-75 text-white">
```
StevenMalaihollo, line 2: I think you should contain this abolsute div in a relative container so that it won't overlap the topbar :D

StevenMalaihollo (approved):

PR #113, opened 2021-06-05
First test by adding a user to a channel

philipdalen:


server/test/app.test.ts
```diff
@@ -29,38 +29,9 @@ describe.skip('Feathers application tests', () => {
   it('starts and shows the index page', async () => {
     const { data } = await axios.get(getUrl());
```
JakubC-projects, line 30: I don't think we have an index page on the server. And this test does not pass for me.

JakubC-projects (approved):

server/test/app.test.ts
philipdalen: done

PR #115, opened 2021-06-07
load answers after reconnect

JakubC-projects:


client/src/components/poll-results.vue
```diff
@@ -96,6 +100,12 @@ export default defineComponent({
                 this.sortedAnswers[answer.answerId].count ++
                 this.totalCount ++
             }
+        },
+        async reloadAnswers(poll: Poll){
+            poll.answers.forEach((option: Option) => {
+                this.sortedAnswers[option.answerId].count = 0;
+            })
+            this.loadAnswers(poll);
```
StevenMalaihollo, line 108: await?

client/src/components/poll-results.vue
```diff
@@ -96,6 +100,12 @@ export default defineComponent({
                 this.sortedAnswers[answer.answerId].count ++
                 this.totalCount ++
             }
+        },
+        async reloadAnswers(poll: Poll){
+            poll.answers.forEach((option: Option) => {
```
StevenMalaihollo, line 105: If I where you I would create an Init method that would run:

createSortedAnswer and loadAnswers from the poll. And then you could run this init method upon create and upon reload right.

JakubC-projects: done that

StevenMalaihollo (approved):

PR #118, opened 2021-06-07
automatically open a poll if a polling event has no polls + rename poll to question

JakubC-projects:


client/src/localization/master.json
```diff
@@ -33,13 +33,13 @@
   "labels": {
```
StevenMalaihollo, line 33: What about publish-poll and these?

StevenMalaihollo (approved):

PR #121, opened 2021-06-07
clickable results and polling events

JakubC-projects:


client/src/components/polling-event-card.vue
```diff
@@ -50,11 +52,11 @@ export default defineComponent({
         }
     },
     methods: {
-        goToLobby(pollingEventKey:string) {
-            this.$router.push({ path: `/polling-event/lobby/${pollingEventKey}`, params: { id: pollingEventKey}})
-        },
-        adminEvent(pollingEventKey:string) {
-            this.$router.push({ path: `/polling-event/admin/${pollingEventKey}`, params: { id: pollingEventKey}})
+        goToEvent() {
+            if (this.$user.personID === this.pollingEvent.creatorId)
```
StevenMalaihollo, line 56: What about if the event is finished? I believe than it shouldn't be clickable

client/src/components/poll-result-tile.vue
```diff
@@ -1,13 +1,12 @@
...
-    <label class="text-gray-700">{{$t('labels.poll-result-visibility.'+poll.resultVisibility)}} - {{votes}} {{$t('labels.votes')}}</label>
-  </div>
+  
+    <router-link :to="`/poll/result/${poll._key}`" class="border rounded-md py-3 px-4 cursor-pointer">
+        <div class="flex justify-between">
+            <h4 class="font-bold">{{poll.title}}</h4>
+            <ArrowRightIcon class="ml-4 w-6 h-6 text-blue-900"/>
+        </div>
+        <label class="text-gray-700">{{$t('labels.poll-result-visibility.'+poll.resultVisibility)}} - {{votes}} {{$t('labels.votes')}}</label>
+    </router-link>
```
StevenMalaihollo, line 9: You need to merge master into this. the page looks like a very old version

JakubC-projects: this is the master version

StevenMalaihollo (approved):

PR #123, opened 2021-06-07
Finished event

JakubC-projects:


client/src/views/AdminEvent.vue
```diff
@@ -4,7 +4,7 @@
             <div class="w-full h-full px-4 py-8">
                 <PollingEventPanel :pollingEvent="pollingEvent" @reloadPollingEvent="loadPollingEvent"/>
                 <div class="flex py-8 gap-6 font-bold justify-center cursor-pointer" :class="inactiveTabColor">
-                    <h3 :class="currentTab === 'polls' ? activeTabColor : ''" @click="currentTab='polls'">{{$t('labels.polls')}}</h3>
```
StevenMalaihollo, line 7: Are you sure they don't want to have the option to restart the polling event?

JakubC-projects: I don't know man

client/src/views/ThankYou.vue
```diff
@@ -0,0 +1,20 @@
...
+            <div class="form-section text-center padding-md">
+            <p class="font-bold text-blue-500 mb-3">
+                {{$t('labels.bcc-vote')}}
+            </p>
+            <h2 class="font-bold mb-3">{{$t('labels.thank-you')}}</h2>
+            <p class="text-gray-700 mb-10 pb-1">
+                {{$t('info.results-will-be-visible')}}
+            </p>
+            <router-link to="/">
+                <button class="bg-gray-200 w-full p-4 rounded-full mb-5">
```
StevenMalaihollo, line 13: I think this button should have a max-width.

It looks a bit too wide on desktop

StevenMalaihollo: Maybe 350px

client/src/views/AdminEvent.vue
```diff
@@ -4,7 +4,7 @@
             <div class="w-full h-full px-4 py-8">
                 <PollingEventPanel :pollingEvent="pollingEvent" @reloadPollingEvent="loadPollingEvent"/>
                 <div class="flex py-8 gap-6 font-bold justify-center cursor-pointer" :class="inactiveTabColor">
```
StevenMalaihollo, line 6: I think that the cursor pointer should be moved to the individual elements

client/src/views/AdminEvent.vue
StevenMalaihollo: Just remove the v-if for now then :) Until the business comes with feedback

client/src/views/ParticipantLobby.vue
```diff
@@ -41,6 +41,8 @@ export default defineComponent({
         
         this.$client.service('poll').on('patched', this.getPoll)
 
+        this.$client.service('polling-event').on('patched', this.patchEvent);
```
StevenMalaihollo, line 44: Wow but now if any polling-event gets patched to inactive the user gets sent to the thankyou page.

Should probably check if it is the correct one too.

client/src/localization/master.json
```diff
@@ -20,6 +20,7 @@
     "join-this-event": "Join this voting event",
     "start-live-poll": "Start this polling event",
     "close-live-poll": "Close this polling event",
+    "restart-live-poll": "restart this polling event",
```
StevenMalaihollo, line 23: Capital R

client/src/views/ParticipantLobby.vue
JakubC-projects: It is not possible because I only get this event if I am in the right channel

client/tailwind.config.js
```diff
@@ -54,6 +54,12 @@ module.exports = {
         },
         blue: {
           300: '#EBF4FF',
+          <<
+          << << < HEAD
+          500: '#758CDF',
+          ===
+          === = >>>
+          >>> > d5daa630b72bb7046b1521bf7d7d744e303ac8ac
```
StevenMalaihollo, line 62: There is an issue here!

StevenMalaihollo (approved):

PR #130, opened 2021-06-07
Socket fix fix 

JakubC-projects:


client/src/views/ParticipantLobby.vue
```diff
@@ -35,17 +35,17 @@ export default defineComponent({
         }
     },
     async created() {
-        
-        this.loadPollingEvent();
-        this.loadCurrentPoll();
+        this.init()
```
StevenMalaihollo, line 38: Are you sure that there is no awaiting needed?

client/src/components/poll-results.vue
```diff
@@ -93,6 +94,7 @@ export default defineComponent({
             }
         },
         async init(){
+            console.log('initialize');
```
StevenMalaihollo, line 97: Console log, and we don't use semicolons in the frontend

StevenMalaihollo (approved):
Merge after removing clog.

Issue #119, opened 2021-06-07
Bugfix User roles using incorrect data


StevenMalaihollo:
**User Story**
As a User , I want to always be able to use the role that I have in Members.

**Additional Description**
Currently due to a bug in Members 2.0 the user roles are using an outdated version of the roles for my user.

**Acceptance Criteria**
1. Adding a role in members should make it usable after logging in to the voting app.

PR #134, opened 2021-06-07
Get roles from related property from members

philipdalen:
https://github.com/bcc-code/bcc-vote/issues/119

### DIFF
--- a/server/src/services/authentication/authentication.ts
+++ b/server/src/services/authentication/authentication.ts
@@ -8,6 +8,7 @@ import { expressOauth, OAuthStrategy, OAuthProfile } from '@feathersjs/authentic
 import { NotAuthenticated } from '@feathersjs/errors';
 import { Application } from '../../declarations';
 import pick from 'lodash/pick';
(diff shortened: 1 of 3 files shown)


server/src/services/authentication/authentication.ts
```diff
@@ -26,20 +27,14 @@ class Auth0Strategy extends OAuthStrategy {
...
-            for (const role of member.roles ) {
-                if (role.scope) delete role.scope;
-                if (role.name) delete role.name;
-                if (role.active !== undefined) delete role.active;
-                if (role.securityLevel) delete role.securityLevel;
-                if (role.administrator !== undefined) delete role.administrator;
-            }
+            member._id = `user/${member._key}`;
+            member.roles = getRolesForPerson(member)
+            member.churchName = member.church.org.name
```
StevenMalaihollo, line 35: Currently the answer.hooks looks at user.church.org.name.

With this change this needs to be refactored

philipdalen: done

StevenMalaihollo (approved):

PR #153, opened 2021-06-08
User identity confirmation popup

JakubC-projects:


client/src/components/polling-event-card.vue
```diff
@@ -1,5 +1,6 @@
 <template>
-    <div class="p-4 border-2 border-gray-200 rounded-lg shadow-base" :class="{'cursor-pointer': !isEventFinished}" @click="goToEvent()">
+    <div>
+    <div class="p-4 border-2 border-gray-200 rounded-lg shadow-base" :class="{'cursor-pointer': !isEventFinished}" @click="showConfirm = true">
```
StevenMalaihollo, line 3: I'd say that the confirm should only be shown if they go to the Lobby. Not the Admin page.

client/src/router/index.ts
```diff
@@ -10,9 +10,10 @@ const router: any = createRouter({
 async function authentication() {
     try {
         const { user } = await router.$client.reAuthenticate()
-        
+
         router.$user._key = user._key
         router.$user._id = user._id
+        router.$user.name = user.displayName
```
StevenMalaihollo, line 16: I think this should just be displayName to keep things consistent.

client/src/components/poll-popover.vue
```diff
@@ -15,7 +15,11 @@
         </div>
         <transition name="fade">
             <div v-if="showConfirm && !hasSavedAnswer">
-                <VoteConfirm :chosenOption="chosenOption" @cancel="showConfirm = false" @confirm="submitAnswer(chosenOption)"/>
+                <VoteConfirm @cancel="showConfirm = false" @resign="showConfirm = false" @confirm="submitAnswer(chosenOption)"><div class="text-center">
+                    <h3 class="font-bold mb-3">{{$t('labels.vote-confirmation')}}</h3>
```
StevenMalaihollo, line 19: Indenting is a bit weird here

StevenMalaihollo:
Could you merge master into here?

StevenMalaihollo (approved):

PR #157, opened 2021-06-09
Admin result page fix

JakubC-projects:


client/src/components/poll-results.vue
```diff
@@ -43,7 +43,8 @@ import { Poll, PollResultVisibility, Answer, Option } from '../domain'
 import { defineComponent, PropType } from 'vue'
 export default defineComponent({
     props: {
-        poll: Object as PropType<Poll>
+        poll: {type: Object as PropType<Poll>, required: true},
+        isAdmin: {type: Boolean, default: false}
```
StevenMalaihollo, line 47: I think that the setting states that only the event creator is allowed to see the answers right? Maybe isEventCreator would be a better name

client/src/components/poll-results.vue
```diff
@@ -61,12 +62,19 @@ export default defineComponent({
...
+            if(this.poll.resultVisibility === PollResultVisibility['Anonymous'])
                 return false
-            }
+            if(this.isAdmin)
+                return true
+                 
+            return false
+        },
+        infoOnVisibility():string{
+            return this.$t('info.poll-is.'+this.poll.resultVisibility)
```
StevenMalaihollo, line 77: Is this computed prop necassary? You could have it straight in the html cant you?

StevenMalaihollo (approved):

PR #167, opened 2021-06-09
New channel tests

JakubC-projects:


server/test/app.test.ts
```diff
@@ -8,32 +8,32 @@ import app from '../src/app';
...
-  after(function(done) {
-    server.close(done);
-  });
-
-  it('starts and shows the index page', async () => {
-    try {
-      const res = await axios.get(getUrl('does/not/exist'));
-      assert.fail("This method should not pass since the page does not exist");
-    } catch (error) {
-      assert.equal(error.response.status,404)
```
StevenMalaihollo, line 34: This test fails for me as status of response is undefined, is it needed?

PR #173, opened 2021-06-10
archiving events + home remake

JakubC-projects:


client/src/views/Home.vue
```diff
@@ -42,31 +66,59 @@ export default defineComponent({
...
+        @apply gap-6;
+        @apply justify-items-stretch;
+    }
     .banner-text{
         height: 190px;
     }
-    @media (min-width: 640px) {
-    .banner-text{
-      height: 310px;
+    @media (min-width: 720px) {
```
StevenMalaihollo, line 114: These breakpoints are pretty random. I believe tailwind and the universal breakpoints for md or lg screens are 768px and 992px respectively

server/src/domain/Poll.ts
```diff
@@ -1,5 +1,5 @@
 export enum PollActiveStatus {
     'Not Started' = 'not_started',
     'Live' = 'live',
-    'Finished' = 'finished'
```
StevenMalaihollo, line 4: Missing archived right

client/src/views/Home.vue
```diff
@@ -42,31 +66,59 @@ export default defineComponent({
...
     },
     methods: {
         goToCreate() {
             this.$router.push({ path: '/create' })
         },
-        async loadMeetings(){
+        async loadPollingEvents(){
             const roleIds = this.$user.roles.map((r: any) => r.id)
-            this.pollingEvents = await this.$client.service('polling-event').find({})
+            const allEvents = await this.$client.service('polling-event').find({})
```
StevenMalaihollo, line 88: unused roleIds var

server/src/domain/Poll.ts
JakubC-projects: The poll doesn't get archived, It's the polling event

client/src/views/Home.vue
JakubC-projects: fixed that 720, the 640 is actually the universal small screen breakpoint

StevenMalaihollo (approved):

PR #174, opened 2021-06-10
answer colors change

JakubC-projects:


server/src/permissions/appAbility.ts
```diff
@@ -69,7 +69,7 @@ const superAdminRoles = ['Developer','CentralAdministrator','SentralInformasjons
 
 export function defineAbilityFor(user:UserDetails, activeRole?:RoleName): AppAbility {
     const builder = new AbilityBuilder<AppAbility>(AppAbility);
-    let abilityRole = activeRole ? activeRole : user.highestLevelRole;
+    let abilityRole = user.activeRole ? user.activeRole : user.highestLevelRole;
```
StevenMalaihollo, line 72: Should pull in master to fix this :D

JakubC-projects: You see it was the pulling from master that caused it in the first place

StevenMalaihollo (approved):

PR #180, opened 2021-06-10
result page changes

JakubC-projects:


client/src/components/poll-results.vue
```diff
@@ -1,33 +1,40 @@
...
+                <span :style="[`background-color: ${sortedAnswers[option.answerId].bgColor};`,`width: ${totalCount === 0 ? 0 : sortedAnswers[option.answerId].count / totalCount * 100}%;`]"
+                    :class="sortedAnswers[option.answerId].count == totalCount ? 'rounded-lg' : 'rounded-l-lg'">
+                    <h5 class="font-bold text-white whitespace-nowrap overflow-hidden ml-12" style="white-space: nowrap;">{{option.label}}</h5>
+                </span>
+                <div class="absolute top-0 flex flex-row-reverse justify-between w-full p-3">
+                    <h5 class="font-bold text-gray-600">
+                        {{sortedAnswers[option.answerId].count}} {{sortedAnswers[option.answerId].count === 1? $t('labels.vote'): $t('labels.votes')}}
+                    </h5>
+                    
+                    <span v-if="chosenOption === option.answerId" class="h-5 w-5 p-1 border border-dark-100 mr-4 bg-dark-100 rounded-full">
```
StevenMalaihollo, line 18: This checkbox isn't showing up for me.

client/src/components/poll-results.vue
```diff
@@ -1,33 +1,40 @@
...
+                <div class="absolute py-3 top-0">
+                    <h5 class="font-bold ml-12" :style="`color: ${sortedAnswers[option.answerId].bgColor}; white-space: nowrap;`">
+                        {{option.label}}</h5>
+                </div>
+                <span :style="[`background-color: ${sortedAnswers[option.answerId].bgColor};`,`width: ${totalCount === 0 ? 0 : sortedAnswers[option.answerId].count / totalCount * 100}%;`]"
+                    :class="sortedAnswers[option.answerId].count == totalCount ? 'rounded-lg' : 'rounded-l-lg'">
+                    <h5 class="font-bold text-white whitespace-nowrap overflow-hidden ml-12" style="white-space: nowrap;">{{option.label}}</h5>
+                </span>
+                <div class="absolute top-0 flex flex-row-reverse justify-between w-full p-3">
+                    <h5 class="font-bold text-gray-600">
```
StevenMalaihollo, line 14: I think that it should be white, but gray when it is not overlapped yet.

client/src/components/poll-results.vue
```diff
@@ -1,33 +1,40 @@
...
             <div v-if="pollResultsAreVisible">
-                <h4 class="font-bold mb-8">{{$t('labels.participants')}}</h4>
+                <h5 class="font-bold mb-3 text-gray-600">{{$t('labels.participants')}}</h5>
                 <transition-group name="list" tag="div">
                     <div v-for="(answer,index) in answers" :key="answer._key">
-                        <div class="py-2 flex justify-between items-center border-gray-200"
+                        <div class="py-3 flex justify-between items-center border-gray-200"
                             :class="index === 0 ? '' : 'border-t-half'">
                             <div>
                                 <h4 class="font-bold">{{answer.displayName}}</h4>
```
StevenMalaihollo, line 32: After removing the churchName it is thus nolonger needed to save it on the answer.

Could you remove it from the answer.hooks method that adds it?

client/tailwind.config.js
```diff
@@ -65,6 +66,10 @@ module.exports = {
           500: '#EB7362',
           600: '#E2345E',
         },
+        dark: {
+          100: 'rgba(0, 0, 0, 0.1)',
+          200: 'rgba(0, 0, 0, 0.15)',
```
StevenMalaihollo, line 71: I wouldn't add these to the tailwind. These are of classes which Tailwind is now generating almost 100 classes for per color for different elements. So to keep the load small we need to make smart choices here

client/src/components/poll-results.vue
```diff
@@ -1,33 +1,40 @@
...
-                    </div>
-                    <span :style="[`background-color: ${sortedAnswers[option.answerId].bgColor};`,`width: ${totalCount === 0 ? 0 : sortedAnswers[option.answerId].count / totalCount * 100}%;`]"
-                        :class="sortedAnswers[option.answerId].count == totalCount ? 'rounded-lg' : 'rounded-l-lg'">
-                        <h5 class="font-bold text-white whitespace-nowrap overflow-hidden ml-12" style="white-space: nowrap;">{{option.label}} ({{sortedAnswers[option.answerId].count}})</h5>
+            <div  v-for="option in poll.answers" :key="option" class="result-bar mb-2">
+                <div class="absolute py-3 top-0">
+                    <h5 class="font-bold ml-12" :style="`color: ${sortedAnswers[option.answerId].bgColor}; white-space: nowrap;`">
+                        {{option.label}}</h5>
+                </div>
+                <span :style="[`background-color: ${sortedAnswers[option.answerId].bgColor};`,`width: ${totalCount === 0 ? 0 : sortedAnswers[option.answerId].count / totalCount * 100}%;`]"
```
StevenMalaihollo, line 9: What about the progress bar Border? It is a slightly darker color in the design. Right now there is a little gap. Could you fix it?

client/src/components/poll-results.vue
```diff
@@ -71,11 +85,16 @@ export default defineComponent({
...
             if(this.isEventCreator)
                 return true
-                 
             return false
-        }
+        },
     },
     methods: {
+        barWidthPercent(opt: Option): number{
+            return this.sortedAnswers[opt.answerId].count / this.totalCount * 100;
```
StevenMalaihollo, line 93: Hmm, I am very unsure about this. Of course this is a lot cleaner. But I can't find any Vue docs about this being supported. This is actually what computed props are for right. For now this is ok, but let me know if you know any docs about using methods in the template

client/src/components/poll-results.vue
```diff
@@ -122,31 +141,33 @@ export default defineComponent({
...
 <style>
-.result-bar {
-  box-sizing: content-box;
-  position: relative;
-  border-width: 1.5px;
-  @apply rounded-lg;
+.dark-ring{
+    box-shadow: inset 0 0 0 2.5px rgba(0, 0, 0, 0.1);
+}
+.calc-width {
```
StevenMalaihollo, line 147: All these class names are very generic, this means that we need to watch out what class names we use in other parts of the project.

We can solve this by making this style scoped or by adding some more conditions to the styling like:
.results-bar .dark-ring {...}
.results-bar .animation {...}

client/src/components/poll-results.vue
JakubC-projects: https://stackoverflow.com/questions/40522634/can-i-pass-parameters-in-computed-properties-in-vue-js

client/tailwind.config.js
StevenMalaihollo: Remove this?

client/src/components/poll-results.vue
JakubC-projects: I fixed that with scopting the style tag, hope that's fine

StevenMalaihollo (approved):

PR #187, opened 2021-06-11
Admin can participate in the event they created

JakubC-projects:


client/src/components/polling-event-card.vue
```diff
@@ -82,6 +80,19 @@ export default defineComponent({
                 date = day + ' , ' + time
             }
             return date
+        },
+        canParticipate():boolean{
+            const rolesEnum: string[] = this.$user.roles.map((r:any) => r.enumName);
+            const filter = this.pollingEvent.participantFilter;
+            if(this.$user.age >= filter.maxAge)
```
StevenMalaihollo, line 87: But the server side permissions should already take care of this? A Poll I am not allowed to participate in should never get sent to the frontend.

But according to the task admins should be able to join them aswell, so that needs to be added to the permissions instead I believe

client/src/components/polling-event-form.vue
```diff
@@ -55,7 +55,7 @@ import { PollingEvent, PollingEventType, PollingEventStatus } from '../domain'
 
 interface Role {
     name: string,
-    _key: string,
+    enumName: string,
```
StevenMalaihollo, line 58: Type RoleName?

StevenMalaihollo (approved):

PR #218, opened 2021-06-15
Added languages support

philipdalen:


crowdin.yaml
```diff
@@ -0,0 +1,11 @@
+"project_id": "260842"                     #open project settings and go to API section 
+"api_token": "ba14916d60835a35156f122c8341ca187b2a71b424a41a1fd77f6327543d4c732fa8f426540314ed"          #open profile settings and go to API & SSO > New Token > create Token
```
StevenMalaihollo, line 2: Is this token a readonly token? otherwise might be smart to replace it with an env variable

StevenMalaihollo (approved):

PR #221, opened 2021-06-15
Result page base on firebase

JakubC-projects:


client/src/assets/style.css
```diff
@@ -200,5 +200,5 @@
   .list-enter-from,
   .list-leave-to {
     opacity: 0;
-    transform: translateY(-30px);
```
StevenMalaihollo, line 203: I liked this animation that it would drop into place instead of just appearing with the opacity like you changed here. Could you run it by Lida what she thinks is better?

client/src/components/poll-form.vue
```diff
@@ -77,12 +77,12 @@ export default defineComponent({
                     {
                         label: this.$t('labels.yes'),
                         explanation: this.$t('explanations.agree'),
-                        answerId: new Date().getTime(),
+                        answerId: (new Date().getTime()).toString(),
```
StevenMalaihollo, line 80: We shouldn't use the dateTime as an answerId. Probably an incremental number is good enough. But keep that for another PR.

client/src/components/poll-popover.vue
```diff
@@ -65,6 +65,7 @@ export default defineComponent({
                     _from: this.poll._id,
                     _to: this.$user._id,
                     answerId: option.answerId,
+                    pollId: this.poll._key,
```
StevenMalaihollo, line 68: I don't think we need this Id. You can extract the key from the _id

client/src/assets/style.css
JakubC-projects: The problem was that it looked weird when changing from selecting one answer to another

client/src/components/poll-popover.vue
JakubC-projects: good point

client/src/components/poll-results.vue
```diff
@@ -65,19 +72,76 @@ export default defineComponent({
...
+            let promises = []
+            promises.push(this.loadBars(this.poll))
+            if(this.selectedOption || this.isEventCreator)
+                promises.push(this.loadAllAnswers(this.poll))
+
+            this.generateSortedOptions();
+            await Promise.all(promises);
+            this.loaded = true
+        },
+        barWidthPercent(opt: Option): number{
```
StevenMalaihollo, line 98: Barwidth and isEndRounded are not used here.

client/src/components/results-progress-bars.vue
```diff
@@ -30,7 +30,9 @@ export default defineComponent({
     props: {
         sortedOptions: {type: Object as PropType<SortedOptions>, required: true},
         totalCount: {type: Number, required: true},
-        chosenOption: {type: Number}
+        chosenOption: {type: Number},
+        modelValue: {type: String, required: true},
+        visibleResults: Boolean,
```
StevenMalaihollo, line 35: I don't think we need to pass the visibleResults prop in here.

Just let this component update the v-model whatever the results Visibility is. And then decide what to do with the result visibility in the place where the actual results are shown.

client/src/components/poll-results.vue
```diff
@@ -65,19 +73,77 @@ export default defineComponent({
...
+            return this.allAnswers.filter((v:Answer)=>{
+                return v.answerId == answerId
+            })
+        },
+        async loadBars(poll:Poll){
+            const pollResults = await this.$client.service('poll-result').get(poll._key).catch(this.$showError)
+            this.changeBars(pollResults);
+        },
+        async loadAllAnswers(poll:Poll){
+            const res = await this.$client.service('answer').find({
```
StevenMalaihollo, line 116: $showError?

JakubC-projects: done

client/src/components/poll-results.vue
JakubC-projects: done

client/src/components/poll-results.vue
```diff
@@ -65,19 +75,73 @@ export default defineComponent({
...
+
+            let promises = []
+            promises.push(this.loadBars(this.poll))
+            if(this.selectedOption || this.isEventCreator)
+                promises.push(this.loadAllAnswers(this.poll))
+
+            await Promise.all(promises);
+            this.loaded = true
+        },
+        getOnlyOneOption(answerId: string): Array<Answer>{
```
StevenMalaihollo, line 101: I don't feel this method is needed as it is only used once. To keep the code clean I would move it to th voterList computed property.

But this is just a preference

client/src/components/poll-results.vue
```diff
@@ -65,19 +75,73 @@ export default defineComponent({
...
+            return this.allAnswers.filter((v:Answer)=>{
+                return v.answerId == answerId
+            })
+        },
+        async loadBars(poll:Poll){
+            console.log('loading bars');
```
StevenMalaihollo, line 107: Forgot some consolelogs

client/src/components/poll-results.vue
```diff
@@ -65,19 +75,73 @@ export default defineComponent({
...
-            const answers = await this.$client.service('answer').find({query}).catch(this.$showError)
-            answers.forEach((a:Answer) => {this.addAnswer(a)})
         },
-        createSortedAnswer(poll:Poll){
+        changeBars(data: PollResult){
+            this.totalCount = 0;
+            for(const ans in data.answerCount){
+                if(data.answerCount.hasOwnProperty(ans)){
+                    console.log(ans);
+                    this.sortedOptions[ans].count = data.answerCount[ans];
```
StevenMalaihollo, line 134: Shouldn't the sortedOptions and totalCount be a computed property? Then we won't have to set it each time the pollResults change

client/src/components/poll-popover.vue
```diff
@@ -4,14 +4,14 @@
             <InfoBox v-if="!hasSavedAnswer" class="mb-4" @closed="infoBoxClosed = true">
```
StevenMalaihollo, line 4: You removed the usage implementation for infoBoxClosed, but it is still referenced here.

Also some unused imports here

client/src/components/results-progress-bars.vue
StevenMalaihollo: Ok, makes sense.

client/src/components/results-voter-list.vue
```diff
@@ -0,0 +1,32 @@
...
+import { Answer, SortedOptions } from '../domain'
+import { defineComponent, PropType } from 'vue'
+export default defineComponent({
+    props: {
+        voterList: {type: Array as PropType<Array<Answer>>, required: true},
+        sortedOptions: {type: Object as PropType<SortedOptions>, required: true},
+    },
+    data(){
+        return {
+            selectedOption: undefined as undefined|number,
```
StevenMalaihollo, line 28: Unused?

client/src/components/admin-polls-panel.vue
```diff
@@ -67,12 +67,6 @@ export default defineComponent({
             }
                 
             if(poll){
-                if(poll.activeStatus === PollActiveStatus['Finished'])
```
StevenMalaihollo, line 70: Why is this done?

JakubC-projects: Now it gets removed in a hook

client/src/components/poll-results.vue
JakubC-projects: I don't see how sortedOptions could be a computed property, but changed the total count to be a computed property

client/src/components/results-voter-list.vue
JakubC-projects: removed

client/src/components/poll-popover.vue
JakubC-projects: done

client/src/components/poll-results.vue
JakubC-projects: done

client/src/components/poll-results.vue
StevenMalaihollo: Look at the vuex store, we need to replace it with vuex in the end anyway but there is an example there already

server/src/services/poll/poll.hooks.ts
```diff
@@ -21,9 +23,50 @@ const addTime = async (context: HookContext):Promise<HookContext> => {
...
+            batchCount = 0;
+        }
+    });
+    await batch.commit();
+};
+
+const removeFromArango = (context: HookContext) => {
+    return context.app.services.answer.remove(null, {
+        query: {
+            _from: context.result._key
```
StevenMalaihollo, line 55: the _from is an _id not a _key

StevenMalaihollo: What about removing the Firestore poll-result document?

JakubC-projects: It doesn't need to be deleted, it gets overwritten

JakubC-projects: done

server/src/channels.ts
```diff
@@ -1,87 +1,101 @@
...
+    app.services['poll-result'].publish('patched', (data: any) => {
         return app.channel(data.pollingEventId);
-      }
+    });
 
+    app.services.poll.publish('patched', async (data:Poll) => {
+
+        if(!data.firestore){
+            data.firestore = true;
+            if(data.activeStatus === PollActiveStatus['Live']){
```
StevenMalaihollo, line 41: Move to a hook

client/src/components/results-progress-bars.vue
```diff
@@ -42,9 +44,17 @@ export default defineComponent({
         },
         isEndRounded(option:SortedOption):boolean {
             return this.barWidthPercent(option) > 97
+        },
+        clickOption(answerId: number):void {
```
StevenMalaihollo, line 48: number?

StevenMalaihollo (approved):

PR #227, opened 2021-06-16
Vote action label fix norwegian

### LINKED ISSUE #223: Crowdin - translation
**User Story**

**Additional Description**
....

**Acceptance Criteria**
1. vote button and vote on result page should be two different translations
2.

StevenMalaihollo:
Closes #223


client/src/localization/es_vote_master.json
```diff
@@ -58,8 +58,8 @@
     "show-results-live": "Show results live",
     "count": "Count",
     "voters": "Voters",
-    "votes": "Votes",
-    "vote": "Vote",
+    "votes": "votes",
```
JakubC-projects, line 61: Shouldn't the "Voters" label be in smaller case too then

JakubC-projects (approved):

client/src/localization/es_vote_master.json
StevenMalaihollo: True, on it

PR #230, opened 2021-06-16
Create report

JakubC-projects:


client/src/components/admin-results-panel.vue
```diff
@@ -20,16 +18,148 @@ import ClipboardListIcon from 'heroicons-vue3/outline/ClipboardListIcon'
...
+        getVoters(){
+            const allVoterSet = new Set();
+            this.allAnswers.forEach((ans: Answer) => {
+                allVoterSet.add(ans._to);
+            })
+            return this.$client.service('user').find({
+                query: {
+                    $limit: allVoterSet.size,
+                    _id: {
+                        $in: [...allVoterSet]
```
StevenMalaihollo, line 130: I think this query won't work on scale. If we want to export a list of 8000 participants I think the list of $in will be to long and not performant at all.

client/src/components/admin-results-panel.vue
```diff
@@ -20,16 +18,148 @@ import ClipboardListIcon from 'heroicons-vue3/outline/ClipboardListIcon'
...
+            answerMap: {} as {[answerId: string]: string},
+            workSheets: {} as {[pollId: string]: any},
+            workBook: {} as any,
+            allAnswers: [] as Answer[],
+            voterMap: {} as {[voterId: string]: any},
+            columnWidth: [10, 40, 5, 20, 30, 20, 20, 20, 30, 20],
+        }
+    },
+    methods: {
+        async generateReport():Promise<void>{
```
StevenMalaihollo, line 43: I think we should try to extract these functions to its own file. There are so many methods that it completely floods this file.

Maybe to a vue mixin or preferably its own vuex module :D

server/src/permissions/appAbility.ts
```diff
@@ -63,6 +64,8 @@ const rolePermissions: Record<string, DefinePermissions> = {
         can('find', 'org');
         can('find', 'role');
         can('find', 'answer');
+        can('find', 'user');
```
StevenMalaihollo, line 67: I think that the voting Admin can only find users he is responsible for? 

The org._id is saved on the votiongAdmin role so we can use that.

client/src/domain/Poll.ts
```diff
@@ -8,6 +8,7 @@ export interface Poll {
   answers: Option[],
   activeStatus: PollActiveStatus;
   resultVisibility: PollResultVisibility;
+  lastChanged: number;
```
StevenMalaihollo, line 11: Date instead of number?

client/src/components/admin-results-panel.vue
JakubC-projects: So what do you propose?

client/src/domain/Poll.ts
JakubC-projects: Those are actually just stored as numbers

StevenMalaihollo: Yes, I just checked. 

Unix timestamps are considered as numbers in Typescript, you are right

server/src/permissions/appAbility.ts
JakubC-projects: Done

client/src/components/admin-results-panel.vue
JakubC-projects: I've just moved the logic to external .ts file, hope that's fine

client/src/components/admin-results-panel.vue
StevenMalaihollo: Add a limit of 50:

- find the first 50
- add them to the list
- get the next 50


And don't forget a loading spinner and showError with such a big request

JakubC-projects: ON it

JakubC-projects: Done that

server/src/permissions/appAbility.ts
```diff
@@ -63,11 +67,16 @@ const rolePermissions: Record<string, DefinePermissions> = {
         can('find', 'org');
         can('find', 'role');
         can('find', 'answer');
+        cannot('find', 'answer', {'visibility': 'anonymous' as any});
```
StevenMalaihollo, line 70: Tests!!!

StevenMalaihollo: And you can't use the PollVisibility type here as the values?

client/src/views/LiveResults.vue
```diff
@@ -38,9 +38,12 @@ export default defineComponent({
     },
     async created(){
         await this.init()
-        this.$client.service('answer').on('created', this.ADD_ANSWER)
+        
         this.$client.service('poll').on('patched', this.patchedPoll)
         this.$client.service('poll-result').on('patched', this.UPDATE_POLL_RESULT)
+        if(this.resultsVisible)
```
StevenMalaihollo, line 44: I think this is a bug:

On created it checks if the current ActivePolls results are visible or not. Then after that a new poll which is visible appears and the results don't get listened to anymore.

I understand the thought process here. But I think that filtering out the hidden answers should be guarded on the server side.

server/src/services/poll/poll.hooks.ts
```diff
@@ -60,6 +60,7 @@ const removeFromArango = (context: HookContext) => {
 const resetPollResults = (context: HookContext) => {
     const pollRes = {
         pollingEventId: context.result.pollingEventId,
+        pollId: context.result._key,
```
StevenMalaihollo, line 63: Do you have a test for reseting the poll Results? You can do it in a seperate PR but please add it.

server/src/domain/Answer.ts
```diff
@@ -1,11 +1,14 @@
...
     _id: string;
     _key: string;
     _from: string;
     _to: string;
     firestore?: boolean;
+    visibility: PollResultVisibility
     answerId: string;
     pollingEventId: string;
     displayName: string;
     churchName: string;
```
StevenMalaihollo, line 13: churchName is not needed anymore right?

client/src/views/LiveResults.vue
JakubC-projects: good point

server/src/domain/Answer.ts
JakubC-projects: removed

server/src/permissions/appAbility.ts
JakubC-projects: Yes, we can!

server/src/services/poll-result/poll-result.class.ts
```diff
@@ -17,8 +23,26 @@ export class PollResult implements ServiceMethods<Data> {
     }
 
     // eslint-disable-next-line @typescript-eslint/no-unused-vars
-    async find (params?: Params): Promise<Data[] | Paginated<Data>> {
-        throw new NotImplemented();
+    async find (params: Params): Promise<Data[] | Paginated<Data>> {
+        const query = params.query;
+        console.log('looking for');
```
StevenMalaihollo, line 28: logs

StevenMalaihollo (approved):

PR #237, opened 2021-06-18
Fix for anonymous votes

adelinn:


adelinn:
We should actually not remove the addChannel hook from pool.hooks.ts for now.

server/src/channels.ts
```diff
@@ -17,14 +19,12 @@ export default function(app: Application): void {
...
-    app.services.answer.publish('created',async (data:Answer) => {
-
+    app.services.answer.publish('created',async (data:Answer, context:any) => {
         if(!data.firestore){
             data.firestore = true;
             await db.collection('answer').doc(data._key).set(data);
-        }else{
-            if(data.visibility !== PollResultVisibility['Anonymous'])
-                return app.channel(data.pollingEventId);
+        }else if(data.visibility==PollResultVisibility['Public'] || data.visibility !== PollResultVisibility['Anonymous'] && context.params.user?.activeRole && adminRoles.includes(context.params.user?.activeRole)){
```
StevenMalaihollo, line 26: This is actually a permissions check right. The same as in the app.hooks file.

Could you change this that instead of being an if statement with many checks, that we use the casl.js $can check?
I think you can look at how we did it in app.hooks aswell. Please reach out if you need help.

server/src/services/answer/answer.hooks.ts
```diff
@@ -12,15 +12,18 @@ const preventMultipleVotes= async (context: HookContext) => {
     return context;
 };
 
-const preventVoteOnInactivePoll = async (context:HookContext) => {
+const preventVoteOnInactivePollAndAddVisibility = async (context:HookContext) => {
```
StevenMalaihollo, line 15: We believe it is cleaner code to have 1 concern per hook function. For example to have a validation or prevention function:

Could you add a seperate function that adds the Visibility to the data?

server/src/channels.ts
adelinn: Ok. Maybe tomorrow.

server/src/services/answer/answer.hooks.ts
adelinn: I thought about that but then I'll have to do the get poll request twice...

server/src/services/answer/answer.hooks.ts
```diff
@@ -13,14 +13,24 @@ const preventMultipleVotes= async (context: HookContext) => {
...
+    
+    context.data.visibility = poll.resultVisibility;
+
+    return context;
+};
+
+const addVisibility = async (context:HookContext) => {
+    const key = context.data._from.split('/')[1];
+    const poll = await db.collection('poll').doc(key).get().data();
+    context.data.visibility = poll.resultVisibility;
```
StevenMalaihollo, line 33: Ahh, now I see the issue.

You are correct. We should keep it in 1 hook for now then. And please merge master into this branch sothat the changes are up to date.

StevenMalaihollo: After these changes we should test this thouroughly right. Deploying to dev using the trigger in the root folder. Please contact me if there is anything wrong

adelinn: I made the changes, but I've not tested this thoroughly.

PR #243, opened 2021-06-18
Back to arango db

### LINKED ISSUE #240: Move back to arangoDB
**Additional Description**
We want to use the firestore database only for real time connection for now.

**Acceptance Criteria**
1. Use firestore only for real time connection

JakubC-projects:
closes #240

### DIFF
--- a/client/src/store/result.ts
+++ b/client/src/store/result.ts
@@ -6,7 +6,8 @@ export interface ResultState {
   pollingEvent: PollingEvent | null,
   polls: Array<Poll>,
   pollResult: PollResult,
-  answers: Array<Answer>
+  answers: Array<Answer>,
+  votersLoadedFromFind: Array<string>
 }
 
 const result: Module<ResultState,RootState> = ({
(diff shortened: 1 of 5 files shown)


client/src/store/result.ts
```diff
@@ -15,14 +16,16 @@ const result: Module<ResultState,RootState> = ({
         pollingEvent: null,
         polls: [],
         pollResult: {} as PollResult,
-        answers: []
+        answers: [],
+        votersLoadedFromFind: []
```
StevenMalaihollo, line 20: I think the naming is important here. This is meant to be an array of answerIds right?

StevenMalaihollo: Voters from find seems a bit odd. :/

JakubC-projects: I've actually been using the user id for that, though I see the answerId's fit better for that function

JakubC-projects: the from find actually means something, because I'm searching only through the answers that come from the initial find request

StevenMalaihollo (approved):

PR #246, opened 2021-06-18
event tile design change

### LINKED ISSUE #198: Design change - buttons on home screen
**User Story**

**Additional Description**
....

**Acceptance Criteria**
1. Update the buttons
2. Tags/Status - Color, size and styling was slightly changed after we changed the main color and added it to the event section
https://www.figma.com/file/Eq4kNcMejzrIQWV6sD4e0C/BCC-Voting?node-id=899%3A751
3. Updated design for the cards: https://www.figma.com/file/Eq4kNcMejzrIQWV6sD4e0C/BCC-Voting?node-id=921%3A138

JakubC-projects:
closes #198


client/src/components/event-filter-info.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-    <div class="flex flex-wrap gap-x-2">
+    <div class="flex flex-wrap gap-x-2 text-xl">
```
StevenMalaihollo, line 2: I addeed the classes for each type of text that Lida defined in the design:

For example text-xl is a label html element and a text lg is a h5 element. So if you see an h5 in the design tab you know straight away to use that element and we save coding like we do here.

client/src/components/polling-event-status.vue
```diff
@@ -0,0 +1,24 @@
+<template>
+    <h4 class="font-bold px-3 py-1 rounded-full flex-shrink-0" :class="styling">{{$t(`labels.polling-event-status.${status}`)}}</h4>
+</template>
+
```
StevenMalaihollo, line 4: Awesome! Really cool plan for a reusable component

client/src/views/AdminEvent.vue
```diff
@@ -2,7 +2,7 @@
     <div :class="backgroundColor">
         <div class="max-w-5xl mx-auto">
             <div class="w-full h-full px-4 py-8">
-                <PollingEventPanel :pollingEvent="pollingEvent" @reloadPollingEvent="loadPollingEvent"/>
+                <PollingEventPanel v-if="isEventLoaded" :pollingEvent="pollingEvent" @reloadPollingEvent="loadPollingEvent"/>
```
StevenMalaihollo, line 5: Spinners if not loaded? May look weird for the user otherwise.

client/src/components/event-filter-info.vue
JakubC-projects: done

client/src/views/AdminEvent.vue
JakubC-projects: done

StevenMalaihollo (approved):

PR #249, opened 2021-06-21
Result page answer count fix

### LINKED ISSUE #245: Live result page counts 1 vote before anyone voted
**User Story**

image

**Acceptance Criteria**
1. Totalt should show 0 as long as nobody voted.

JakubC-projects:
closes #245


client/src/store/result.ts
```diff
@@ -104,9 +104,6 @@ const result: Module<ResultState,RootState> = ({
             return sortedOptions
         },
         answerCount: (state:ResultState):number => {
-            if(!state.pollResult){
```
StevenMalaihollo, line 107: This will throw an error if it is used in an unexpected way. Because it will try to use the answerCount of an undefined object. 

Or goal should be to make the functions as pure as possible right. I think removing this check is degrading that

JakubC-projects: Ok, though the pollResult always exists at least as an empty object which is actually truthy

JakubC-projects: so maybe that check should check answerCount property

JakubC-projects: Done

client/src/store/result.ts
```diff
@@ -3,99 +3,94 @@ import { Poll, PollingEvent, PollActiveStatus, Answer, SortedOptions, Option, Po
...
 }
 
+
 const result: Module<ResultState,RootState> = ({
     namespaced: true,
     state: ():ResultState => ({
-        pollingEvent: null,
-        polls: [],
-        pollResult: {} as PollResult,
         answers: [],
```
StevenMalaihollo, line 17: Readd undefined types

client/src/store/result.ts
```diff
@@ -3,99 +3,94 @@ import { Poll, PollingEvent, PollActiveStatus, Answer, SortedOptions, Option, Po
...
-                        updatedPollsArray.push(poll)
-                    }
-                });
-            }
-            commit('UPDATE_POLLS',updatedPollsArray)
+            if(!state.pollingEvent || patchedPoll.pollingEventId !== state.pollingEvent._key)
+                return;
+            if(patchedPoll.activeStatus === PollActiveStatus['Live'])
+                commit('UPDATE_ACTIVE_POLL',patchedPoll)
+            else
```
StevenMalaihollo, line 65: add condition for checking if patchedPoll is finished and same as previous activePOll

StevenMalaihollo (approved):

PR #250, opened 2021-06-21
disable removing notifications on click

### LINKED ISSUE #222: Sometimes a participant doesn't get up the error message "can't vote 2x"
**User Story**

**Additional Description**
....

**Acceptance Criteria**
1.  Sometime participant doesn't get up the error message "can't vote 2x"

JakubC-projects:
I'm not exactly sure what causes this issuel My guess is that if the error shows in the same place as the voting button user may accidentally close the notification. It is possible to happen and I don't know how could it happen otherwise, so I suppose that was the cause.

closes #222


client/src/main.ts
```diff
@@ -27,7 +27,7 @@ const app = createApp(App)
 
 app.component('Spinner', Spinner)
 app.component('InfoBox', InfoBox)
-app.use(Toast,{duration:2000, positionX: 'right',positionY:'bottom'})
+app.use(Toast,{duration:2000, positionX: 'right',positionY:'bottom', disableClick: true})
```
StevenMalaihollo, line 30: I think on mobile it needs to be added to the top of the screen instead of the bottom.

JakubC-projects: done

StevenMalaihollo (approved):

PR #256, opened 2021-06-21
fix to the double vote issue

### LINKED ISSUE #254: Votes displayed multiple times
During a recent live event, voter names were shown multiple times in the list of votes.

In this instance, some voters were shown once, whereas others were shown 2, 3 or 4 times. The issues affected multiple, but not all voters - but was visible on all user devices.

35 voters were online, and the issue occurred only for one question (of roughly 5)

User Story
In Drammen a member could vote twice:

https://projects.bcc.no/desk/tickets/6898616/messages

Acceptance Criteria

Each member should only be able to vote once

JakubC-projects:
closes #254

### DIFF
--- a/server/src/services/answer/answer.hooks.ts
+++ b/server/src/services/answer/answer.hooks.ts
@@ -2,19 +2,7 @@ import { HookContext } from "@feathersjs/feathers";
 import { Answer, PollActiveStatus } from "../../domain";
 import { db, FieldValue } from '../../firestore';
 
-const preventMultipleVotes= async (context: HookContext) => {
-    const vote = await context.app.service('answer').find({
-        query: {
-            _from: context.data._from,
-            _to: context.data._to,
-        }
-    });
-    if(vote.length > 0)
-        throw Error('You cannot vote 2 times');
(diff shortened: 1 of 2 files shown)


server/src/services/answer/answer.hooks.ts
```diff
@@ -38,31 +26,45 @@ const addUserData = async (context:HookContext) => {
...
+    const pollKey = context.data._from.split('/')[1];
+    context.data._key = pollKey+'-'+context.params?.user?._key;
+
+    return context;
+};
+
+const incrementCounter = async (context:HookContext):Promise<HookContext> => {
+    const { result } = context;
+    const pollKey = result._from.split('/')[1];
     const pollRef = db.collection('poll-result').doc(pollKey);
```
StevenMalaihollo, line 39: By changing the pollkey this doc id would be incorrect then right.

JakubC-projects: I don't understand. I'm not changing the pollKey anywhere

StevenMalaihollo (approved):

PR #260, opened 2021-06-23
Login domain change

### LINKED ISSUE #253: Users asked if they want to open link in BrunstadTV app
Android users are asked if they want to open BCC Vote in Chrome or BrunstadTV app when they click on a link.

This happens when users are not logged in and are redirected to bcc-sso.

The underlying reason is that the BrunstadTV app also subscribes to the bcc-sso domain. A quick workaround is to use login.bcc.no as the auth domain instead of the bcc-sso domain. The long-term solution is to remove associations to auth from the BrunstadTV app (or assign a dedicated auth domain to the BTV app).

The problem is not isolated to BCC Vote, but is highly critical since it can effectively make it impossible to users to access BCC Vote (if they choose "always open" in BrunstadTV).

### LINKED ISSUE #261: Change auth0 domain to login.bcc.no
**Additional Description**
We need to use login.bcc.no because otherwised users are asked to open the website in brunstadTv app

**Acceptance Criteria**
1. When user logs in he gets redirected to login.bcc.no instead of bcc-sso.eu.auth0.com

JakubC-projects:
closes #253

### DIFF
--- a/server/src/app.hooks.ts
+++ b/server/src/app.hooks.ts
@@ -1,3 +1,4 @@
+import logger from './logger';
 import { HookContext } from "@feathersjs/feathers";
 import { defineAbilityFor,  } from './permissions/appAbility';
 import {authenticateExternal } from './services/authentication/customAuthentication';
@@ -191,7 +192,8 @@ const purgeErrors = (context: HookContext):void => {
 };
 
 const logErrors = (context: HookContext):void => {
-    console.error(context.error.name ,'from app.hooks.ts:', context.error.message,'during', context.method, context.path);
+    const message = [context.error.name, 'from app.hooks.ts:', context.error.message, 'during', context.method, context.path].join(' ');
+    logger.info(message);
 };
 
 export default {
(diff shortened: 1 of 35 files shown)


StevenMalaihollo:
Any testing possible for the logic you changed?

JakubC-projects:
> 
> 
> Any testing possible for the logic you changed?

I don't think there is beside e2e tests

server/src/app.hooks.ts
```diff
@@ -1,229 +1,227 @@
...
+        const requestedAction = context.method;
+        const requestedSubject:any = context.path;
+
+        let results;
+        if(Array.isArray(context.result)){
+            results = context.result;
+        }else{
+            results = context.result.data;
+        }
+        const allowedItems = [];
```
StevenMalaihollo, line 116: Can it still push new items to this array even though it is a const? In my head a const won't change after declaration

JakubC-projects: Because It's still the same array.

StevenMalaihollo:
> Of course you can 

I think that this usage of const is a bad practise, the whole purpose of const and let/var is that const can work as a kindaof readonly property which I can trust the value. 

In my opinion a bit of a design mistake in JS, but it works so its: approved

StevenMalaihollo (approved):

PR #267, opened 2021-06-24
Add sentry io

### LINKED ISSUE #264: Add sentry io
**User Story**
As an administrator, I want to see an overview of the errors that users encounter

**Acceptance Criteria**
1. Connect the frontend to sentry.io

JakubC-projects:
closes #264 
It works surprisingly well because it will automatically send all the unhandled errors

### DIFF
--- a/client/src/main.ts
+++ b/client/src/main.ts
@@ -41,9 +41,9 @@ if(window.location.hostname === 'vote.bcc.no'){
         config: {id: 'G-6V21WXD03F'}
     })
     
+    init({dsn: 'https://de460cd536b34cdab822a0338782e799@o879247.ingest.sentry.io/5831770'})
 }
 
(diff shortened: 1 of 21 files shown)


client/src/main.ts
```diff
@@ -47,6 +48,14 @@ if(window.location.hostname === 'vote.bcc.no'){
 app.mixin({
     methods: {
         $showError(error: Error) {
+            const personID = app.config.globalProperties.$user?.personID
+            if(personID)
+                setTag('personID', personID)
+            if(error.message.includes('Validation Error') || error.message === 'You cannot vote 2 times')
```
StevenMalaihollo, line 54: There should be another way to determine if an error is just informative or really a problem that needs to be fixed. Maybe we should just send all $showErrors from the backend as info messages. Maybe you have some other idea.

Would be cool to talk about it and maybe get a showcase from you 👍🏽

StevenMalaihollo (approved):

PR #270, opened 2021-06-28
Create feedback page

### LINKED ISSUE #265: Feedback page
**User Story**

**Additional Description**
at the end of a vote the user should be asked to give some feedback
Part of the last page

**Acceptance Criteria**
1. one textbox to answer  "did you experience any trouble?"  with submit button
2. optional rating 1-5 
3. users can skipp the feedback
4. follow design: https://www.figma.com/file/Eq4kNcMejzrIQWV6sD4e0C/BCC-Voting?node-id=1111%3A186

JakubC-projects:
closes #265


client/src/components/feedback-popover.vue
```diff
@@ -0,0 +1,115 @@
...
+    overflow: auto;
+    height: 70vh;
+    margin-top:30vh;
+}
+
+.centering-div {
+    @apply w-6;
+}
+
+@media screen and (min-width: 768px) {
```
StevenMalaihollo, line 101: All this popover dialog logic and styling is duplicated here. Could you create a reusable component? I think we need it in several places, like the confirm popover.

client/src/components/input-textarea-field.vue
```diff
@@ -0,0 +1,56 @@
...
+    <textarea v-model="model" ref="textArea" @change="setHeight" @keydown.enter.prevent="clickEnter()">
+    </textarea>
+</template>
+
+<script lang="ts">
+import { defineComponent } from 'vue'
+
+export default defineComponent({
+    props: {
+        modelValue: {type: String, required: true},
```
StevenMalaihollo, line 11: Baseheight as a string?

client/src/router/routes.ts
```diff
@@ -33,16 +33,16 @@ export const routes: Array<RouteRecordRaw> = [
...
+        name: 'Thank you',
+        path: '/thank-you',
         component: () => require('../views/ThankYou.vue'),
         meta: {
             bgColor: "bg-blue-900"
         }
     },
     {
         name: 'Poll result',
-        path: '/poll/result/:id',
```
StevenMalaihollo, line 45: Why is the id of the poll removed? This way you could link and send it to another admin in the future.

client/src/components/input-textarea-field.vue
```diff
@@ -0,0 +1,56 @@
...
+        baseHeight: {type: String, required: true},
+        enableNewLine: {type: Boolean, required: true}
+    },
+    data(){
+        return{
+            paddingAndBorderVertical: 19.2
+        }
+    },
+    computed: {
+        textHeight():void{
```
StevenMalaihollo, line 21: Unused

client/src/router/routes.ts
JakubC-projects: Whoops, that was accidental, fixed

client/src/components/input-textarea-field.vue
JakubC-projects: done

server/src/services/feedback/feedback.class.ts
```diff
@@ -0,0 +1,84 @@
...
+interface Data {
+    pollingEventId: string,
+    personID: string,
+    rating: number,
+    message?: string
+}
+
+const possibleRatings = [1, 2, 3, 4, 5];
+
+const updateRating = async(rating: number, eventId: string):Promise<void> => {
```
StevenMalaihollo, line 17: Remember that we are using Firestore as a just an eventstore. Not a datastore. I think an arango document should be updated, but please call me if we should discuss it.

client/src/components/input-textarea-field.vue
JakubC-projects: Well, that makes some sense, this way you can pass here things like 1rem or 50% and it will work

client/src/components/feedback-popover.vue
JakubC-projects: I guess you're right. I'll fix that tomorrow

server/test/feedback-tests.ts
```diff
@@ -0,0 +1,22 @@
...
+
+describe('polling-event', async () => {
+    let testSet:any;
+
+    before(async ()=>{
+        testSet = pollingEventsTestSet();
+    });
+    it('Create feedback', async () => {
+        try{
+            const data = testSet.feedbackData();
```
StevenMalaihollo, line 14: No asserts on how the feedback result should be? This isn’t testing much then 😸

JakubC-projects: Well, i don't know how to assert anything here because it is coming from the firestore. And it is checking if there are no errors happening, so that's something

server/src/services/feedback/feedback.class.ts
JakubC-projects: Philip said it can be like that for now

client/src/components/feedback-popover.vue
JakubC-projects: Done

StevenMalaihollo (approved):

PR #280, opened 2021-06-30
Report fix

### LINKED ISSUE #277: Report can't be downloaded
**User Story**
voting admin from Oslo is trying to download the export is only loading 

https://projects.bcc.no/desk/tickets/6931053/messages

**Acceptance Criteria**
1. Should be able to download the report of a voting event

root cause: 9c038c53d001ba6ae9ddc86e59045b9c03605d43

JakubC-projects:
closes #277 

I wasn't able to find the real reason why it happened or reproduce this behavior without external manipulation of the DB. I was actually able to generate the report for this event without any problems.

In the sentry (https://sentry.io/organizations/bcc-hj/issues/2483366193/?project=5831770&query=is%3Aunresolved) we can check that for some reason one or more users weren't properly retrieved from the db and I have no idea how that could have happened.

I made a "fix" so that if the user data is not present then the report will just not display any information about that person. 

And BTW I mixed in here some small changes to the feedback logic, sorry for that.


StevenMalaihollo (approved):

StevenMalaihollo:
Uncaught (in promise) TypeError: Value sent to Number function of cells ["C5"] was not a number, it has type of undefined and value of undefined
    at cellBlock.numberSetter [as number] (index.js?c47e:77)
    at eval (generateReport.ts?49b7:73)
    at Proxy.forEach (<anonymous>)
    at fillPollInfo (generateReport.ts?49b7:70)
    at eval (generateReport.ts?49b7:58)
    at Proxy.forEach (<anonymous>)
    at createSheets (generateReport.ts?49b7:57)
    at generateReport (generateReport.ts?49b7:16)
    at _callee$ (admin-results-panel.vue?328c:37)
    at tryCatch (runtime.js?96cf:63)

I get this error when trying to generate a report.

client/src/components/admin-results-panel.vue
```diff
@@ -49,7 +54,7 @@ export default defineComponent({
             const allVoters = await this.getVoters(allAnswers)
             const allResults = await this.getResults()
 
-            const excelFile = generateReport(this.pollingEvent, this.savedPolls, allAnswers, allVoters, allResults)
+            const excelFile = generateReport(this.pollingEvent, this.startedPolls, allAnswers, allVoters, allResults)
```
StevenMalaihollo, line 57: $handleError

StevenMalaihollo (approved):

PR #282, opened 2021-06-30
VoteAdmin with correct Role orgs

### LINKED ISSUE #189: All voting admins for the same church should be able to access, administrate polling events of other voting admins for the same church
**User Story**
e.g. if Lene is voting admin for Oslo and Morten is also voting admin for Oslo. Than they both should be able to see & administrate the voting events the other created even if they are not in the defined user group of the voting event

**Acceptance Criteria**
1. Administrators of the same church can access & administrate the poll events of the other administrators of the same church

StevenMalaihollo:
Closes #189


JakubC-projects (commented):
This actually fails 8 tests. I think it's because the test data wasn't updated, so I guess we should update that before releasing.

server/src/domain/User.ts
```diff
@@ -1,12 +1,29 @@
...
+    email: string;
+    roles: Array<UserRole>;
+    displayName: string;
+    age: number;
+    cellPhone?: any;
+}
+
+export type UserRole = Role & {
+    organisationID: Array<number>;
+    org_id: Array<string>;
```
JakubC-projects, line 18: Why we have both organisationID and org_id? I see that org_id is never used, so I think it should be removed and organisationID would then be renamed to orgID or something (it feels wierd to use the '_' notation here).

server/src/permissions/appAbility.ts
```diff
@@ -53,21 +53,22 @@ const rolePermissions: Record<string, DefinePermissions> = {
...
+        const votingAdminFor = user.roles.filter((r:UserRole) => r.enumName == 'VotingAdmin')[0].organisationID;
         can('create','poll');
         can('update', 'poll');
         can('remove', 'poll');
         can('patch', 'poll');
 
-        can('patch', 'polling-event', { 'participantFilter.org': userChurchID });
-        can('update', 'polling-event', { 'participantFilter.org': userChurchID });
-        can('create','polling-event', { 'participantFilter.org': userChurchID });
+        can('patch', 'polling-event', { 'participantFilter.org': { $in: votingAdminFor}} as any);
```
JakubC-projects, line 63: Why do we use $in if votingAdminFor is just a number?

StevenMalaihollo: votingAdminFor is an array of numbers like defined in the types

server/src/domain/User.ts
StevenMalaihollo: organisationID is our external ids array right. And org_id is the internal arango db id.

We shouldn't use the org_id in vote (external system) but we have this array on the UserRole so we might aswell have it in the types right.

server/src/permissions/appAbility.ts
JakubC-projects: Oh, sorry It really is, but I guess we should rename the organisationID to something that would indicate that it's an array

server/src/domain/User.ts
JakubC-projects: But we could remove that array. It doesn't have to exist, does it? For  me it doesn't feel right to store unnecessary data especially that it has a similar name to used data

StevenMalaihollo: No it is necassary, thats the whol point of this PR: You can be a votingAdmin for multiple churches

JakubC-projects: I understand now that organisationID is an array and it is necessary but the org_id isn't, am I right?

StevenMalaihollo: Yes

JakubC-projects (approved):

PR #283, opened 2021-07-02
Poll activity

### LINKED ISSUE #275: the poll is still visible even thought it is not active anymore
**User Story**

**Additional Description**

https://projects.bcc.no/desk/tickets/6930877/messages

**Acceptance Criteria**
1. when the question is "closed" it shouldn't show the question anymore

root cause: 7f013a03fd2d03854c586151f50f38288b708f9d

JakubC-projects:
closes #275

The actual fix is on the 62nd line of Lobby.vue
It was caused by the fact that we didn't update the current poll if the loadCurrentPoll method hadn't found any active polls


elpenny (approved):
Looks correct and better than earlier version. Hopefully this should fix mentioned bug.

PR #288, opened 2021-07-05
Type errors - part 1

### LINKED ISSUE #285: Sentry errors - TypeError
**Additional Description**
In the sentry io we see:
https://sentry.io/organizations/bcc-hj/issues/?project=5831770

**Acceptance Criteria**
1. Remove that error

JakubC-projects:
closes #285 

it's supposed to fix those:
https://sentry.io/organizations/bcc-hj/issues/2483484218/?project=5831770&query=is%3Aunresolved
https://sentry.io/organizations/bcc-hj/issues/2483368059/?project=5831770&query=is%3Aunresolved

### DIFF
--- a/client/src/components/input-textarea-field.vue
+++ b/client/src/components/input-textarea-field.vue
@@ -21,16 +21,12 @@ export default defineComponent({
         model: {
             get():string{
                 // timeout is so that the height change happens after the text is rendered
-                setTimeout(() => {
-                    const element = this.$refs.textArea as HTMLFormElement
-                    element.style.height = this.baseHeight
-                    element.style.height = (element.scrollHeight) +'px'
-                }, 0)
+                this.adjustHeight()
                 
                 return this.modelValue
             },
             set(newVal:string){
-                this.$emit('update:modelValue', newVal);
+                this.$emit('update:modelValue', newVal)
             }
         }
     },
@@ -38,6 +34,15 @@ export default defineComponent({
         clickEnter(){
             if(this.enableNewLine)
                 this.model += '\n'
+        },
+        adjustHeight(){
+            setTimeout(() => {
+                const element = this.$refs.textArea as HTMLFormElement
+                if(!element)
+                    return
+                element.style.height = this.baseHeight
+                element.style.height = (element.scrollHeight) +'px'
+            }, 0)
         }
     },
     emits: ['update:modelValue']
--- a/client/src/components/poll-results.vue
+++ b/client/src/components/poll-results.vue
@@ -44,7 +44,7 @@ export default defineComponent({
         }
     },
     async created(){
-        this.generateSortedOptions();
+        this.generateSortedOptions()
 
         await this.init()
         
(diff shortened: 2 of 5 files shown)


client/src/components/input-textarea-field.vue
```diff
@@ -21,23 +21,28 @@ export default defineComponent({
...
     },
     methods: {
         clickEnter(){
             if(this.enableNewLine)
                 this.model += '\n'
+        },
+        adjustHeight(){
+            setTimeout(() => {
+                const element = this.$refs.textArea as HTMLFormElement
+                if(!element)
```
JakubC-projects, line 41: This solves one of the errors

client/src/views/ThankYou.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-    <div class="max-w-5xl mx-auto" >
+    <div v-if="!noParams" class="max-w-5xl mx-auto">
```
JakubC-projects, line 2: this solves the other error

client/src/router/routes.ts
```diff
@@ -4,54 +4,54 @@ export const routes: Array<RouteRecordRaw> = [
     {
         path: '/',
         name: 'Home',
-        component: async() => require('../views/Home.vue'),
+        component: () => import('../views/Home.vue' as string),
```
JakubC-projects, line 7: This probably should be in another PR, but the way it works it is allows for lazy loading of the views, so those components would get loaded only when actually necessary, thus allowing the homepage to load faster

client/src/components/poll-results.vue
```diff
@@ -44,7 +44,7 @@ export default defineComponent({
         }
     },
     async created(){
-        this.generateSortedOptions();
+        this.generateSortedOptions()
```
JakubC-projects, line 47: This file is just some linting and small refactor

client/src/router/routes.ts
JakubC-projects: The 'as string' is because my editor was underlining it without that. The code works without it too.

StevenMalaihollo: How is this different from using the require function? I thought this already does what you described here.

Do you have some documentation to back this?

JakubC-projects: https://router.vuejs.org/guide/advanced/lazy-loading.html#grouping-components-in-the-same-chunk

JakubC-projects: The require statement loads the code synchronously and import function actually returns a promise.

client/src/components/poll-results.vue
```diff
@@ -120,30 +120,28 @@ export default defineComponent({
...
         },
         changeBars(data: PollResult){
             if(data.pollId !== this.poll._key)
                 return
-            for(const ans in data.answerCount){
-                if(data.answerCount.hasOwnProperty(ans)){
-                    this.sortedOptions[ans].count = data.answerCount[ans];
-                }
-            }
+            Object.keys(this.sortedOptions).forEach((ans: string) =>{
```
StevenMalaihollo, line 142: I think the old version is more reliable. What if the poll-result change event comes in before the sortedOptions are generated

StevenMalaihollo (approved):
Pr looks good, just the one changeBars function change that I don't really understand

PR #299, opened 2021-07-24
Polling event with Role visible to all

### LINKED ISSUE #298: Events created for a specific role is visible for everyone
**User Story**
I created an event "TEsting Roles" that I limited to one specific role "central administrator" I asked Lidiya to test it - she has the roles "pmo manger", "developer" and "member" in Members . According to that she shouldn't get up this event

**Acceptance Criteria**
1. Events that are selected "all churches" but limited to a specific role should be only available for that role

StevenMalaihollo:
Closes #298

It has become more clear that our permissions system doesn't work with that much of an 'activeRole' concept like it does in Members. The permissions code look at all the roles the use has and decides the allowence based on that.

Because of this I have chosen to change permissions tests in order to test more closely to the reality of the app usage, in for example app.hooks.


JakubC-projects (approved):
Looks good

PR #348, opened 2022-02-09
Add filter for initial data sync

JakubC-projects:
Containes a few changes:
1. Adds lastChanged to poll-result collection
2. Refactors poll-result class
3. Adds a query when registering firestore listeners, so we don't get the whole collection on the initial load.

I suspect that the initial loading of whole collection was the cause of the bug described in #349 though I wasn't able to confirm that. That change will be beneficial eitherway as it's going to reduce startup time and firestore costs (by about 200kr per month)

I added no tests for 2 reasons:
1. There are no new features, the app should work exactly the same way,  just a bit faster.
2. All the changes are related to firestore which cannot be reliably tested unless we create a dev environment there


elpenny (changes requested):
1. build is failing for now.
2. please add some test to cover changed functionality.

server/src/channels.ts
```diff
@@ -1,103 +1,76 @@
...
     }
     
     app.services['polling-event'].publish('patched',async (data:PollingEvent) => {
-        if(!data.firestore){
-            data.firestore = true;
-            db.collection('polling-event').doc(data._key).set(data);
-        }else{
-            return app.channel(data._key);
-        }
+        if(!data.firestore)return;
```
elpenny, line 13: space before return

server/src/channels.ts
```diff
@@ -1,103 +1,76 @@
...
     app.services.poll.publish('patched', async (data:Poll) => {
-
-        if(!data.firestore){
-            data.firestore = true;
-            
-            await db.collection('poll').doc(data._key).set(data);
-        }else{
-            return app.channel(data.pollingEventId);
-        }
+        if(data.firestore) return;
```
elpenny, line 30: wrong conditional

server/src/domain/PollResult.ts
```diff
@@ -0,0 +1,7 @@
+export interface PollResultDetails {
+    answerCount: Record<string, number>;
+    lastChanged: number;
+    pollId: string;
+    pollingEventId: string;
+    firestore: boolean;
+}
```
elpenny, line 7: missing EOF line

server/src/domain/index.ts
```diff
@@ -1,4 +1,5 @@
 export * from './User';
 export * from './Poll';
 export * from './PollingEvent';
-export * from './Answer';
\ No newline at end of file
+export * from './Answer';
+export * from './PollResult';
```
elpenny, line 5: missing EOF line

elpenny (approved):

server/src/channels.ts
JakubC-projects: fixed

server/src/domain/PollResult.ts
JakubC-projects: fixed

server/src/domain/index.ts
JakubC-projects: fixed

PR #370, opened 2022-03-28
Sentry errors fixes

### LINKED ISSUE #369: Vote increase in frontend errors
**User Story**
According to Sentry our crash free sessions have increased drastically.

image

image

It looks like all of our users the past week experience the shown bugs.

**Acceptance Criteria**
1. The bugs that occur so often should be resolved.

StevenMalaihollo:
Closes #369

Add website title

### DIFF
--- a/client/src/store/index.ts
+++ b/client/src/store/index.ts
@@ -17,13 +17,17 @@ export const store: any = createStore<RootState>({
     actions: {
         async findPollingEvents({commit},archived) {
             const query = {
-                
                 status: archived ? {} : {
                     $ne: PollingEventStatus.Archived
                 }
             };
-            const unarchivedEvents = await store.$client.service('polling-event').find({ query});
-            commit('UPDATE_POLLING_EVENTS',unarchivedEvents);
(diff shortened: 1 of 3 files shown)


client/src/store/index.ts
```diff
@@ -17,13 +17,17 @@ export const store: any = createStore<RootState>({
...
             const query = {
-                
                 status: archived ? {} : {
                     $ne: PollingEventStatus.Archived
                 }
             };
-            const unarchivedEvents = await store.$client.service('polling-event').find({ query});
-            commit('UPDATE_POLLING_EVENTS',unarchivedEvents);
+            let currentEvents:PollingEvent[] = [];
+            await store.$client.service('polling-event').find({ query})
```
JakubC-projects, line 25: I'm not sure what was the actuall error here, so I'll that something bad happens if the reuqest fails so we try to UPDATE_POLLING_EVENTS with undefined, Isn't it better then to use try catch here, and have more explicit behavior in case of error?

JakubC-projects (approved):

Issue #377, opened 2022-05-27
Event hidden despite age limit


StevenMalaihollo:
**User Story**
As an 18 year old user , I want to join an event where the age limit is set to 18.

**Additional Description**
https://projects.bcc.no/desk/tickets/10397644/messages

**Acceptance Criteria**
1. Users of the exact age should be able to join the event.

PR #378, opened 2022-05-27
Event min- and max-age filter wasn't working as expected

StevenMalaihollo:
Description: 
Added tests, fixed permissions logic.

Closes https://github.com/bcc-code/bcc-vote/issues/377

### DIFF
--- a/server/src/permissions/appAbility.ts
+++ b/server/src/permissions/appAbility.ts
@@ -21,8 +21,8 @@ const globalPermissions = (user: User, { can, cannot }: AbilityBuilder<AppAbilit
     can('get', 'poll-result');
     can('create', 'feedback');
     can(['find','get'],'polling-event', {'participantFilter.org': {$in: [user.churchID.toString(), 'all']}, 'participantFilter.role': { $in: [...userRoleNames, 'all'] }} as any);
-    cannot(['find','get'],'polling-event',{'participantFilter.minAge': {$gte:user.age}} as any);
-    cannot(['find','get'],'polling-event',{'participantFilter.maxAge': {$lte:user.age}} as any);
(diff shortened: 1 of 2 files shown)


server/src/permissions/appAbility.ts
```diff
@@ -21,8 +21,8 @@ const globalPermissions = (user: User, { can, cannot }: AbilityBuilder<AppAbilit
     can('get', 'poll-result');
     can('create', 'feedback');
     can(['find','get'],'polling-event', {'participantFilter.org': {$in: [user.churchID.toString(), 'all']}, 'participantFilter.role': { $in: [...userRoleNames, 'all'] }} as any);
-    cannot(['find','get'],'polling-event',{'participantFilter.minAge': {$gte:user.age}} as any);
-    cannot(['find','get'],'polling-event',{'participantFilter.maxAge': {$lte:user.age}} as any);
```
JakubC-projects, line 25: Shouldn't this rule stay unchanged. I think the rule should work like  ```minAge <= user.age < maxAge```

server/test/polling-event-tests.ts
```diff
@@ -26,7 +28,42 @@ describe('polling-event', async () => {
...
 
+            event.participantFilter.maxAge = user.age;
+            const eventSubject = subject("polling-event", event);
 
+            const ability = defineAbilityFor(user);
+            assert.equal(ability.can('find', eventSubject), false);
+            assert.equal(ability.can('get', eventSubject), false);
 
+            user.age--;
+            const abilityTooOld = defineAbilityFor(user);
```
JakubC-projects, line 62: I don't think this variable name makes sense

JakubC-projects (approved):

PR #392, opened 2022-07-12
Event replace filtering on ChurchID with role OrgID

StevenMalaihollo:
Updated permissions sothat you can only join an event if you have correct Role and Org. This is totally disconnected from the ChurchID that a member has. Also some UI updates to replace Church with Organization.

Closes https://github.com/bcc-code/bcc-membership/issues/2242


JakubC-projects:
I think we should make the Member role selectable when you create a polling event, because right now it's not possible to create it only for members of a given church. Also I'm not sure if the "all" option for role selection makes sense anymore.

JakubC-projects (approved):

PR #399, opened 2022-08-01
Sanitize sheet name

JakubC-projects:
removes illegal characters from the sheet names so that they can be properly displayed in excel


client/src/functions/generateReport.ts
```diff
@@ -111,3 +111,7 @@ const getResultMap = (allResults: any[]) => {
     });
     return resultMap;
 };
+
+function sanitizeSheetName(name: string) {
+    return name.replace(/[\\/*[\]]/g, "");
```
philipdalen, line 116: Replace with underscore
What about sheet length limitation?

JakubC-projects: fixed

PR #407, opened 2022-08-04
Use-rest-in-vote

### LINKED ISSUE #385: Vote - verify solution works

JakubC-projects:
1. Uses rest client in vote
2. Adds http load test
3. Fixes a misconfiguration that prod instance would connect to dev firebase instance

closes #385


client/src/client.ts
```diff
@@ -1,22 +1,10 @@
...
 import hooks from "./hooks";
-import socketioClient from "@feathersjs/socketio-client";
-import io from 'socket.io-client';
+import restClient from "@feathersjs/rest-client";
 
 export const client = feathers();
-const socket = io(window.location.hostname === 'localhost' ? 'http://localhost:4040' : `${location.origin}`, {
-    transports: ['websocket', 'polling'],
-});
+const rest = restClient(window.location.hostname === 'localhost' ? 'http://localhost:4040' : `${location.origin}`);
```
philipdalen, line 6: Can't you use the config object that you get based on the environment here?

philipdalen (approved):

client/src/client.ts
JakubC-projects: done

PR #413, opened 2022-09-13
Hide feedback page

philipdalen:


client/src/views/ThankYou.vue
```diff
@@ -17,9 +17,9 @@
                 </router-link>
             </div>
         </div>
-        <transition name="fade">
+        <!-- <transition name="fade">
```
StevenMalaihollo, line 20: Shouldnt we just remove this code? The history is in Git anyway. and the component isn't removed either

StevenMalaihollo (approved):

PR #414, opened 2022-09-13
Add Dutch as language, finish i18n implementation

StevenMalaihollo:
Closes [#2849](https://github.com/bcc-code/bcc-membership/issues/2849)

Added translation files, Added server logic to fetch the users cultureCode1. Added logic to lazy load only the needed translation messages.

The default locale is Norwegian when no supported locale is found. And if a translation is missing we take it from the master file (which is English).


JakubC-projects (approved):
ziet er goed uit, goedgekeurd

PR #418, opened 2022-11-02
Admin collaboration

StevenMalaihollo:
Closes https://github.com/bcc-code/bcc-membership/issues/2901


server/src/permissions/appAbility.ts
```diff
@@ -103,9 +94,18 @@ const rolePermissions: Record<VoteRoleName, DefinePermissions> = {
             visibility: PollResultVisibility['Non Public'],
         });
         can('find', 'user', {churchID: {$in: votingAdminFor}});
+
+        can(['get', 'find'], 'polling-event', {
+            'participantFilter.org': {$in: votingAdminFor},
+        });
+        can(['find', 'get'], 'polling-event', {creatorId: user.personID});
+
         can('get', 'polling-event-report', {
```
JakubC-projects, line 103: isn't this permission obsolete now?

server/src/permissions/appAbility.ts
```diff
@@ -76,16 +72,11 @@ const rolePermissions: Record<VoteRoleName, DefinePermissions> = {
...
-        });
+        can(['find', 'get'], 'polling-event');
     },
     VotingAdmin(user, {can}) {
         const votingAdminFor = user.roles.filter((r: UserRole) => r.enumName === 'VotingAdmin')[0].orgIDs;
-        can('create', 'poll');
-        can('update', 'poll');
-        can('remove', 'poll');
-        can('patch', 'poll');
+        can('manage', 'poll');
```
JakubC-projects, line 79: This looks like an excessive permission

JakubC-projects:
I think we should consider having the admin panel update in real time, because it may be used by multiple people at the same time

JakubC-projects (approved):

server/src/permissions/appAbility.ts
StevenMalaihollo: No this is exactly the same as it was. Before the VotingAdmins could allready do all the actions

server/src/permissions/appAbility.ts
StevenMalaihollo: Right, true

StevenMalaihollo:
> I think we should consider having the admin panel update in real time, because it may be used by multiple people at the same time

Agreed, will do so in the next PR.

PR #420, opened 2022-11-03
Add realtime events for admins

StevenMalaihollo:
Make frontend reactive for different collaborating admins.

Closes https://github.com/bcc-code/bcc-membership/issues/3142


client/src/store/index.ts
```diff
@@ -33,6 +34,15 @@ export const store = createStore<RootState>({
 
             commit('UPDATE_POLLING_EVENTS', currentEvents);
         },
+        async pollingEventActivation({state, commit}, activation:PollingEventActivation) {
+            const {pollingEvents} = state
+            const pollIndex = pollingEvents.findIndex(p => p._key === activation._key);
+            if(pollIndex < 0) {
+                throw Error(`Could not find activated ${activation._id}`)
```
JakubC-projects, line 41: I don't think this error is correct, it will trigger when a polling event that I don't have access to gets activated

JakubC-projects: I think it would be nice to also trigger this function on creation of ```PollingEventActivation```. We could try to fetch the whole poll from the server, and if we succeed then it would show in the menu.

client/src/views/Home.vue
```diff
@@ -99,12 +106,26 @@ export default defineComponent({
...
         async loadPollingEvents(archived = false) {
             this.loading = true;
             await this.findPollingEvents(archived);
             this.loading = false;
         },
+        async registerEvents(): Promise<void> {
+            try {
+                const pollingEventId = this.pollingEvents[0]?._key
+                if(pollingEventId) {
+                    const collections = await firestore(pollingEventId);
```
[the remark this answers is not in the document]
JakubC-projects: I think we could update ```firebase-credentials``` service to allow for creating tokens without the pollingEventId, that token would then be used to get updates on the polling events.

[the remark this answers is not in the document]
StevenMalaihollo: True, fixed

client/src/store/index.ts
StevenMalaihollo: Good idea! FIxed

PR #421, opened 2022-11-04
Redesign colors and fonts to match BCC styleguide

StevenMalaihollo:
https://github.com/bcc-code/bcc-membership/issues/3108 among others..


JakubC-projects:
image

Are we sure about those green buttons here? I don't think they were updated in the design, but I don't think they fit really well

client/src/components/poll-results.vue
```diff
@@ -47,8 +47,8 @@ export default defineComponent({
             selectedOption: '',
             allAnswers: [] as Array<Answer>,
             answerIdsFromFind: new Set() as Set<string>,
-            answerColors: ['#004C78', '#006887', '#0081A2', '#329BBD', '#55B6D9', '#72D0E3'],
-            neutralColor: '#C1C7DA',
+            answerColors: ['#003934', '#4E9A94', '#C8CCD2', '#329BBD', '#55B6D9', '#72D0E3'], // colors?
```
JakubC-projects, line 50: I don't think this is correct, if there are 4 or more options then the 3rd and the last option are both grey

JakubC-projects: image

JakubC-projects (approved):

client/src/components/poll-results.vue
StevenMalaihollo: Right, there was a bug that the neutral color was in there twice

PR #428, opened 2022-11-16
Feathers global debouncing, Poll and Event button disabling

### LINKED ISSUE #424: Event was created mulitiple times
**Describe the bug**

**To Reproduce**
image

**Expected behavior**

root cause: *commit_id*

StevenMalaihollo:
Closes #424


client/src/hooks/debounce.ts
```diff
@@ -0,0 +1,25 @@
+import {HookContext} from '@feathersjs/feathers';
+
+const cache: Record<string, boolean> = {}
```
JakubC-projects, line 3: I would personally use a set here
```typescript
const cache = new Set<string>()
```

JakubC-projects (approved):

Issue #455, opened 2023-01-19
Participant list


amelienbawolski:
**User Story**
As a .... , I want to .... so that ....

**Additional Description**
....

**Acceptance Criteria**
1.

amelienbawolski:
Participant count with name: 2 days

StevenMalaihollo:
Missing a design here, should it be something like this?

![Image](https://github.com/bcc-code/bcc-vote/assets/39584351/4a5d4ee7-2439-4249-bbe2-7cea91f1755d)

Also is the list a seperate page?

amelienbawolski:
@StevenMalaihollo should be added to the existing report as a seperate tab

amelienbawolski:
<img width="488" alt="image" src="https://github.com/bcc-code/bcc-vote/assets/107836832/1972c6e7-0a5a-45d9-b4db-066a2e136a32">
Could be something like this:
<img width="350" alt="image" src="https://github.com/bcc-code/bcc-vote/assets/107836832/afbb61f9-c465-4e09-b81a-2a452e4f604d">

PR #473, opened 2023-05-24
Count participants

JakubC-projects:
Registers a person as participant when entering the lobby page.
User has to be a participant in order to:
1. Create answers
2. Get firestore token

half of #455

Will create the past participants with query:
```sql
for p in polling_event
    let u = (for a in answer
        filter a.pollingEventId == p._key
        collect user = a._to
        aggregate timestamp = min(a.lastChanged)
        insert {
            _key: concat(p._key, "-", split(user, "/")[1]),
            _from: user,
            _to: p._id,
            timestamp: timestamp
        } into participant
        return NEW)
    return u
```


server/src/domain/Participant.ts
```diff
@@ -0,0 +1,3 @@
+import {DatabaseEdgeEntity} from './Database';
+
+export type Participant = DatabaseEdgeEntity<'participant', 'user', 'polling_event'>;
```
StevenMalaihollo, line 3: It's probably insightful to also add the data for when the participant started.

Like startDate or validFrom

StevenMalaihollo (approved):

StevenMalaihollo:
local test data should be added for the participants collection

server/src/domain/Participant.ts
JakubC-projects: good point added as `timestamp`

PR #476, opened 2023-05-24
Add-firebase-emulator

JakubC-projects:
It was difficult to test the solution locally because of conflicting ids in firebase dev instance.
because of this I needed to add this emulator setup


client/src/firebase/index.ts
```diff
@@ -18,7 +18,7 @@ interface EventCollections extends PublicCollections {
     ['answer']: firebase.firestore.CollectionReference<Answer>;
 }
 
-type Collections = EventCollections | PublicCollections
+type Collections = EventCollections | PublicCollections;
```
StevenMalaihollo, line 21: https://www.reddit.com/r/Firebase/comments/jygigo/firebase_emulator_production_warning/

StevenMalaihollo (approved):

client/src/firebase/index.ts
JakubC-projects: fixed

PR #480, opened 2023-05-25
Show questions for Finished event

### LINKED ISSUE #462: Questions should be still visible after voting event is set to finished
**User Story**
image

Endring av status ikke aktiv og aktiv bør kunne overstyres manuelt slik at man kan tilbakestille en synlig poll til å ikke vises for sluttbruker.

**Acceptance Criteria**
1. Questions should be still visible after the event is set to finished

StevenMalaihollo:
Closes #462


JakubC-projects:
I think we can remove the logic that redirects the user to the result tab after finishing the event

client/src/components/saved-poll.vue
```diff
@@ -112,14 +110,22 @@ export default defineComponent({
...
     data() {
         return {
             changeConfirmation: '',
         };
     },
     computed: {
+        isEventLive(): boolean {
+            return PollingEventStatus.Live === this.eventStatus;
+        },
+        isEditable(): boolean {
```
JakubC-projects, line 124: I would rename this to something like `isEventNotFinished`

JakubC-projects: because I think it's a bit confusing that you can have `isEditable = true` but still cannot edit the poll (because it's been started)

JakubC-projects (approved):

client/src/components/saved-poll.vue
StevenMalaihollo: Fixed

PR #484, opened 2023-05-26
Reset polling event

JakubC-projects:
Adds a button for resetting a polling event


client/src/components/admin-polling-event-panel.vue
```diff
@@ -83,6 +100,22 @@
...
                 </ConfirmPopover>
+                <ConfirmPopover
+                    v-else-if="resetConfirmation"
+                    cancelTranslation="cancel"
+                    confirmTranslation="yes-continue"
+                    @resign="resetConfirmation = false"
+                    @cancel="resetConfirmation = false"
+                    @confirm="resetPollingEvent"
+                >
+                    <template #header>
```
StevenMalaihollo, line 111: I think it would make sense that this Confirmation would be it's own component

server/src/services/polling-event/polling-event.hooks.ts
```diff
@@ -36,6 +37,36 @@ const addFeedbackDocument = async (context: HookContext): Promise<HookContext> =
     return context;
 };
 
+async function handleEventReset(context: HookContext): Promise<HookContext> {
+    const update = context.data as Partial<PollingEvent> | undefined;
+    if (update?.status !== PollingEventStatus['Not Started']) return context;
+    if (typeof context.id !== 'string') return context;
+
+    await resetPolls(context.app, context.id);
```
StevenMalaihollo, line 45: Should probably add a test for this.

client/src/components/admin-polling-event-panel.vue
```diff
@@ -181,6 +219,20 @@ export default defineComponent({
...
+            this.resetConfirmationLoading = true;
+            client
+                .service('polling-event')
+                .patch(this.$route.params.id, {
+                    status: PollingEventStatus['Not Started'],
+                })
+                .then(() => {
+                    this.resetConfirmation = false;
+                    this.resetConfirmationLoading = false;
+                    this.$emit('reloadPollingEvent');
```
StevenMalaihollo, line 232: This `reloadPollingEvent` is does not reload the polls. So if a poll was reset it will keep that state.

StevenMalaihollo:
image
This view looks a bit odd. I know it makes sense in the example bellow. But with only 1 button it looks very assymmetrical. Can't we align it in the center if there is only 1 button?

image

client/src/components/admin-polling-event-panel.vue
JakubC-projects: I'm not sure if helps much, I'd rather have it consistent with the other popup

server/src/services/polling-event/polling-event.hooks.ts
JakubC-projects: Added

client/src/components/admin-polling-event-panel.vue
JakubC-projects: Well, it actually does get updated, because of firestore live updates

JakubC-projects:
> image
This view looks a bit odd. I know it makes sense in the example bellow. But with only 1 button it looks very assymmetrical. Can't we align it in the center if there is only 1 button?

image
 

I don't think the primary button should be moved like that, for me that would seem like a design inconsistency

StevenMalaihollo (approved):

PR #485, opened 2023-05-26
Update the confirm popover

JakubC-projects:
Updates the identity confirmation dialog according to the design.

Additonally fixes a bug that the dialog is not shown if user joins the polling event via a link directly to the lobby. Now the user will get redirected back to the home page with the identity confirmation open


client/src/localization/vote_master.json
```diff
@@ -97,7 +97,11 @@
         "sure-close-event": "Are you sure that you want to close this event?",
         "church": "Church",
         "age": "Age",
-        "role": "Roles"
+        "role": "Roles",
+        "confirm-identity": {
+            "i-am": "I confirm that I am:",
```
StevenMalaihollo, line 102: Need to add the translations for the other languages. I can provide the Dutch ones if needed.

StevenMalaihollo:
Conflicts

StevenMalaihollo:
image

I think the text should be different based on the norwegian that amelie provided:

_I, Philly Daly, hereby confirm that I want to participate in Yearly Meeting Stavanger, and I agree to my data about the polling event being processed._

_Not you? Please sign out_

(Go to login page) should probably say Sign out?     --- (Okay, continue) should probably be Confirm

client/src/functions/storage.ts
```diff
@@ -0,0 +1,8 @@
+export function checkConfirmedIdentity(user_key: string, pollingEvent_key: string) {
```
StevenMalaihollo, line 1: Perhaps `hasConfirmedIdentity` is a better name?

client/src/functions/storage.ts
```diff
@@ -0,0 +1,8 @@
...
+    const savedState = sessionStorage.getItem(`confirm-identity-${user_key}-${pollingEvent_key}`);
+    return savedState !== null;
+}
+
+export function saveConfirmedIdentity(user_key: string, pollingEvent_key: string) {
+    const savedState = sessionStorage.setItem(`confirm-identity-${user_key}-${pollingEvent_key}`, 'true');
```
StevenMalaihollo, line 7: Unused

client/src/localization/vote_master.json
JakubC-projects: Makes sense, would appreciete if you provided the Ducth translations

client/src/functions/storage.ts
JakubC-projects: removed

client/src/functions/storage.ts
JakubC-projects: renamed

PR #486, opened 2023-05-29
Delete polling evnet  after 12 moths

JakubC-projects:


server/src/services/polling-event-cleanup/polling-event-cleanup.service.ts
```diff
@@ -1,24 +1,41 @@
...
-    const apiKey = app.get('authentication')?.archiveEventsApiKey ?? '';
+    const apiKey = app.get('authentication')?.cleanupEventsApiKey ?? '';
     const providedKey = req.headers['x-api-key'];
     return providedKey === apiKey;
 }
 
+async function deleteEvents(app: Application) {
+    const events = await getEventsToBeDeleted(app);
+    for (const event of events) {
+        await app.services['polling-event'].remove(event._key, {});
```
StevenMalaihollo, line 26: Maybe we should add a custom log. Like "Removed polling event 12345678 due to expiry", just to provide a bit of context to the logs.

PR #487, opened 2023-06-01
Timer for Polling event

### LINKED ISSUE #453: Add a timer on the event
**User Story**
Det bør utvikles en funksjon som synliggjør for deltakere og møteansvarlige hvor mange som er innlogget på løsningen i appen (live), slik at ordstyrer har grunnlag for å si at majoriteten har votert.
**Additional Description**

**Acceptance Criteria**
1. add timer on top when a question is published
2. add participant count (of all possible participants)
3. Design:
https://www.figma.com/file/Eq4kNcMejzrIQWV6sD4e0C/01.00-BCC-Voting?type=design&node-id=2013-8319

4. correct design for timer: 
image

StevenMalaihollo:
1/2 of #453


client/src/components/timer.vue
```diff
@@ -0,0 +1,62 @@
+<template>
+    <div class="flex gap-x-1 opacity-0 transition-all ease-in-out duration-300" :class="{'opacity-100': show}">
+        <div class="flex flex-col items-center gap-y-1">
+            <div class="timer-box">
+                <label> {{ elapsedMinutes }} </label>
```
JakubC-projects, line 5: I think minutes should also be padded to always be 2-digit

client/src/components/timer.vue
```diff
@@ -0,0 +1,62 @@
...
+            </div>
+            <p class="caption">Min</p>
+        </div>
+        <span class="mt-1">:</span>
+        <div class="flex flex-col items-center gap-y-1">
+            <div class="timer-box">
+                <label> {{ elapsedSeconds < 10 ? `0${elapsedSeconds}` : elapsedSeconds }} </label>
+            </div>
+            <p class="caption">Sec</p>
```
JakubC-projects, line 14: This should probably be translated

client/src/components/timer.vue
```diff
@@ -0,0 +1,62 @@
...
+            maxElapsedSeconds: 60 * 60,
+            elapsedMinutes: 0,
+            elapsedSeconds: 0,
+            show: false,
+        };
+    },
+    mounted() {
+        setInterval(() => {
+            this.updateElapsedTime();
+        }, 1000);
```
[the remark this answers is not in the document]
JakubC-projects: We may actually need to think about the fact that device clocks may not be synchronised.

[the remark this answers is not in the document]
JakubC-projects: But aside from that we should at least have the timer synced if the device clocks are synced
For that you could consider those solutions:
1. Reduce the interval time (if we set it to e.g. 50 ms, then maximum time out of sync will be 50 ms)
2. Make sure the interval fires at the same time for each user - so basically to make sure that the interval is fired on whole seconds, not in between seconds.

client/src/components/timer.vue
```diff
@@ -0,0 +1,62 @@
...
+    },
+    methods: {
+        updateElapsedTime() {
+            const currentTime = Date.now();
+            const elapsed = Math.floor((currentTime - this.sinceDate) / 1000);
+            if (elapsed >= this.maxElapsedSeconds) {
```
JakubC-projects, line 41: I"m not sure about this one, why does it need to disappear when it reaches 1 hour

client/src/components/timer.vue
```diff
@@ -0,0 +1,62 @@
...
+<script lang="ts">
+import {defineComponent} from 'vue';
+export default defineComponent({
+    props: {
+        sinceDate: {type: Number, required: true},
+    },
+    data() {
+        return {
+            maxElapsedSeconds: 60 * 60,
+            elapsedMinutes: 0,
```
JakubC-projects, line 27: I would presonally only store the elapsedTime (in ms) in the data, and then have those minutes and seconds as calculated properties

client/src/components/timer.vue
```diff
@@ -0,0 +1,62 @@
+<template>
+    <div class="flex gap-x-1 opacity-0 transition-all ease-in-out duration-300" :class="{'opacity-100': show}">
```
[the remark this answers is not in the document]
StevenMalaihollo: Oh, I wasn't here is no time to count down from defined either

client/src/components/timer.vue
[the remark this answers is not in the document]
StevenMalaihollo: Decreased the interval to `100` ms, I think it being out of sync for just 100ms is no problem.

client/src/components/timer.vue
StevenMalaihollo: Fixed

client/src/components/timer.vue
JakubC-projects: Why not? I think it's more correct to display for example 82 minutes there, than to have the clock disapear.

JakubC-projects (approved):

PR #488, opened 2023-06-05
Add participant counter

### LINKED ISSUE #453: Add a timer on the event

StevenMalaihollo:
Closes #453


firebase/firestore.rules
```diff
@@ -8,6 +8,9 @@ service cloud.firestore {
     match /answer/{answer_id} {
       allow read: if request.auth != null && resource.data.visibility == "non_public" && request.auth.token.admin == true && resource.data.pollingEventId == request.auth.token.pollingEventId;
     }
+    match /participant/{participant_id} {
```
JakubC-projects, line 11: Remember that that we don't have a workflow (yet?) to push those changes to the production firestore instance, it will need to be done manually

client/src/components/participation-counter.vue
```diff
@@ -0,0 +1,22 @@
...
+export default defineComponent({
+    props: {
+        participants: {type: Number, required: true},
+        possibleParticipants: {type: Number, required: true},
+    },
+    computed: {
+        percentage(): number {
+            if (!this.possibleParticipants) return 0;
+            const raw = (this.participants / this.possibleParticipants) * 100;
+            return Math.round(raw);
```
JakubC-projects, line 18: I think it should always be rounded down, then if it says 100% it will always mean that everyone actually voted

JakubC-projects:
Wasn't the participant count also suposed to be added in the live-results page https://github.com/bcc-code/bcc-vote/pull/488

server/src/services/index.ts
```diff
@@ -8,13 +8,15 @@ import firebaseCredentials from './firebase-credentials/firebase-credentials.ser
 
 import pollResult from './poll-result/poll-result.service';
 import feedback from './feedback/feedback.service';
+import pollingEventCount from './polling-event-count/polling-event-count.service';
```
JakubC-projects, line 11: Is this endpoint really needed? couldn't the admin just make a find request on in the participants service?

client/src/components/participation-counter.vue
StevenMalaihollo: Good point

server/src/services/index.ts
StevenMalaihollo: Good point, I will refactor this in a next PR.

PR #496, opened 2023-06-08
Add Column Names to each question sheet

### LINKED ISSUE #494: Report should include column names
Currently there is no naming of the columns:

image

**Acceptance Criteria**
1.Report should include column names to make it better understandable for admin
image

StevenMalaihollo:
Closes #494


client/src/functions/generateReport.ts
```diff
@@ -1,8 +1,17 @@
...
+    Answer,
+    PollResult,
+    PollingEvent,
+    PollResultVisibility,
+} from '../domain';
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 import {Workbook, Worksheet} from 'excel4node';
 
 const columnWidth = [10, 40, 5, 30, 30, 20, 20, 20, 30, 20];
+const isBold = {font: {bold: true}};
```
JakubC-projects, line 14: could rename this variable with for example `boldStyle`. `isBold` feels like it would be a boolean

JakubC-projects (approved):

PR #519, opened 2023-07-10
Update-date-picker

### LINKED ISSUE #514: When creating an voting Event, date changes to year 2000
## Describe the bug
When creating an event with the date set to 2023 but **no time is set** date changes to 2000 after creating an event 

image
image

Note: When I put the time, it looks fine 

## Expected behavior
Date shouldn't change to 200 after creating an event. It should stay as it was set even if the time is not set. 

root cause: *commit_id*

JakubC-projects:
# Change summary
1. Updates the date picker in vote

## Change type

-   [ ] No review
-   [x] Small PR
-   [ ] Big PR
-   [ ] Refactor

**Closes #514**


client/src/components/polling-event-form.vue
```diff
@@ -79,7 +79,8 @@
     </div>
 </template>
 
-<script lang="ts">
+<script lang="ts">import InputDateField from './input-date-field.vue';
```
StevenMalaihollo, line 82: Unused import?

client/src/domain/PollingEvent.ts
```diff
@@ -6,7 +6,7 @@ export interface PollingEvent extends DatabaseEntity<'polling_event'> {
     description: string;
     type: PollingEventType;
     status: PollingEventStatus;
-    startDateTime: Date;
+    startDateTime: string;
```
StevenMalaihollo, line 9: The `PollingEvent.ts` file in the server still had the `Date` type. This inconsistency is probably not wanted

StevenMalaihollo (approved):

client/src/components/polling-event-form.vue
JakubC-projects: Man, I'm too used to wriging goLang where all those imports are managed automatically for you

client/src/domain/PollingEvent.ts
JakubC-projects: fixed

PR #520, opened 2023-07-10
fix editing participant filters

### LINKED ISSUE #515: Voting group setting changes when editing it
## Describe the bug
1.How I see my voting event - membership type 
image
2. I go to edit this event 
3. Only "members" is marked 
image
4. I change for example only age and save it 
image
5. It changes also membership types 
image

So if the admin wants to change/edit anything in the voting event, he has to always also set the members type because it changes automatically back to only "members" 
## Expected behavior
Settings should stay the same if going to editing the event
root cause: 3661b83b989301e1cec3c8035acf0ab3c2ac4c91

JakubC-projects:
# Change summary

## Change type

-   [ ] No review
-   [x] Small PR
-   [ ] Big PR
-   [ ] Refactor

**Closes #515**


StevenMalaihollo:
image
Shouldn't it be either the Role _or_ the membership one? 

Currently it's possible to display both

StevenMalaihollo (approved):

PR #523, opened 2023-07-10
Update participatn permission error

### LINKED ISSUE #521: Person cannot access the voting event even if the person is in the target group
## Describe the bug
I logged in as a person 41818 - 17 y.o from Ustron but I get an error when trying to participate the voting event 

image
image

Person is in the target group 
image

## Expected behavior
Person should be able to access the voting event if in the target group. 

root cause: *commit_id*

JakubC-projects:
# Change summary
The incident was caused by lacking permission to find participants (when checking if user is already a participant)
Reverts back the change that was made to prevent trying to cretate a participant multiple times, but without logging the error each time

## Change type

-   [ ] No review
-   [x] Small PR
-   [ ] Big PR
-   [ ] Refactor

**Closes #521**


server/src/services/participant/participant.hooks.ts
```diff
@@ -75,6 +75,7 @@ function handleDuplicateParticipantError(context: HookContext) {
     if (err.errorNum !== UniqueConstraintArangoErrorNum) {
         return context;
     }
+    context.error = undefined;
```
StevenMalaihollo, line 78: Probably add a test to make sure it doesn't happen again

client/src/views/Lobby.vue
```diff
@@ -66,12 +66,7 @@ export default defineComponent({
                 _from: this.$user._id,
                 _to: event._id,
             };
-            const query = {...participant, $limit: 1};
-            const existing = (await client.service('participant').find({query})) as Participant[];
-
-            if (existing.length === 0) {
-                await client.service('participant').create(participant);
-            }
+            await client.service('participant').create(participant);
```
StevenMalaihollo, line 69: I liked the old way of doing it better. Because a create call is heavier on the db than a read call. I think we should tread as lightly as we can on the db.

StevenMalaihollo: I haven't gone to deep into arango's performance, but I think just checking for a unique key before creation is already a heavy operation.

server/src/services/participant/participant.hooks.ts
JakubC-projects: I'm not sure we can really test whether a log was created or not

client/src/views/Lobby.vue
JakubC-projects: I guess we could benchmark it, but I don't see any reason why a failed create call, would be heavier on the db than a read call.

StevenMalaihollo (approved):

PR #547, opened 2023-11-07
Fix-removing-polls

### LINKED ISSUE #546: Can't delete questions
## Describe the bug
1. Go to vote
2. click on the bin simbol to delete a quesiton
3. Can't delete the question get an error message instead

Image

https://projects.bcc.no/desk/tickets/10535237/messages

...
## Expected behavior

- [ ] the vote admin should be able to delete a question

root cause: *commit_id*

JakubC-projects:
# Change summary
1. Fixes a bug that polls are not removed from the firestore

## Change type

-   [ ] No review
-   [x] Small PR
-   [ ] Big PR
-   [ ] Refactor

**Closes #546**


server/test/poll-tests.ts
```diff
@@ -0,0 +1,62 @@
...
+                pollingEventId: event._key,
+                resultVisibility: 'public',
+                title: "This poll has custom explanations. Nice, isn't it?",
+            },
+            {}
+        )) as Poll;
+
+        await app.service('poll').remove(poll._key, {});
+
+        const answers = (await app.service('answer').find({query: {_from: poll._id}})) as Answer[];
```
StevenMalaihollo, line 46: You just created this poll, so it's impossible for there to be answers right? Wouldn't it be more realistic to have an existing poll with answers, and then assert that one to be cleaned up correctly.

StevenMalaihollo (approved):

